PR #107, opened 2014-02-21
Change some defaults to more reasonable values

attila-lendvai:
these shouldn't be too controversial.

the key bindings in the connection list are currently very dangerous when using with production systems, e.g. killing the connected lisp is bound to a mere C-k. the function name ```slime-quit-connection-at-point``` is also misleading, but i didn't touch that (i used to think that it's a disconnect, and restarted production systems until i realized why the users complain about lost web sessions).


attila-lendvai:
anything wrong with this? shall i remove setting the default faces?

i would especially like to see the connection key changes in the official branch. early in our days when we started to operate a live system i kept on killing the inferior lisps unintentionally by pressing 'k' on the connection line, which i thought merely disconnects. well, it did disconnect...

IMO, both the key shortcut is confusing, and the function name also didn't help me. they should be something like `slime-connection-at-point/quit-inferior-lisp` instead of `slime-quit-connection-at-point`.

luismbo:
@attila-lendvai looking at other elisp projects, I see that people inherit from standard faces, in a quasi-semantic way. Here's what I came up with. Does it look decent to you?

```elisp
(define-sldb-faces
  (topline        "the top line describing the error"
                  '(:inherit font-lock-string-face))
  (condition      "the condition class"
                  '(:inherit font-lock-warning-face))
  (section        "the labels of major sections in the debugger buffer"
                  '(:inherit header-line))
  (frame-label    "backtrace frame numbers"
                  '(:inherit shadow))
  (restart-type   "restart names."
                  '(:inherit font-lock-keyword-face))
  (restart        "restart descriptions"
                  '(:inherit font-lock-type-face))
  (restart-number "restart numbers (correspond to keystrokes to invoke)"
                  '(:bold t))
  (frame-line     "function names and arguments in the backtrace")
  (restartable-frame-line
   "frames which are surely restartable"
   '(:foreground "lime green"))
  (non-restartable-frame-line
   "frames which are surely not restartable")
  (detailed-frame-line
   "function names and arguments in a detailed (expanded) frame")
  (local-name     "local variable names"
                  '(:inherit font-lock-variable-name-face))
  (local-value    "local variable values")
  (catch-tag      "catch tags"
                  '(:inherit highlight)))
```

I think I'll remove `slime-face-inheritance-possible-p` since the `:inherit` attribute has been supported since 21.1 (released in 2001). :-)

luismbo:
@attila-lendvai Ever since 82a58ffde0592d47053e73cb1b03b66566c94e49, `slime-net-send` is hard-coded to always use UTF-8, so I believe that renders your commit 2357f9f moot. But, it raises the question: what's slime-net-coding-systems for? (I've removed it here https://github.com/luismbo/slime/commit/5e56df92473a731085ae8c1ced37b5219fa5d092 and it passes the test suite, whatever that's worth.) Your change to swank-compile-file* makes me uneasy though. UTF-8 is not a great fallback encoding because it can lead to decoding errors. I'm open to counterarguments. :-)

attila-lendvai:
@luismbo, re faces: i'm perfectly fine with your ```define-sldb-faces``` form. all i wanted is to have some non-default values, so that people see that there are faces at all.

re utf-8: things probably evolved since my patches. i don't see any changes to ```swank-compile-file``` in the above patches...? my aim was to use utf-8 by default as the quasi standard, so that if a person opens a file with some non-ascii chars in this 21st century we are in, then it just works.

re potential utf-8 errors: i'd say people should mark the encoding of their files if they use some exotic encoding, and we should make sure that the encoding related error messages are clear and helpful, maybe even with some useful emacs "restarts" to install/extend a modeline.

attila-lendvai:
re the remaining two patches:

if your patches reach master, then i can just reopen a new pull request (or change this) with whatever remained. especially because i think the function names are confusing, and i'd like to give them more obvious names besides binding them on a key binding that is harder to press accidentally:

```
slime-connection-at-point/quit-inferior-lisp
slime-connection-at-point/restart-inferior-lisp
slime-connection-at-point/disconnect
```

luismbo:
I meant `read-file` rather than `swank-compile-file`. You mention "exotic encodings"; latin-1 is not that exotic an encoding is it? Opening such a file using UTF-8 will lead to errors even if only comments are non-ascii. And yes, things have evolved since you first wrote that patch. :-)

Actually, that function is overriding the implementation's default external format -- e.g. `sb-impl::*default-external-format*` -- isn't it? That doesn't sound like a good idea... I'm tempted to remove the `(find-external-format "latin-1")` actually. It'd achieve the behaviour you're seeking wouldn't it? (Assuming your Lisp's default external format is UTF-8, of course.)

Your changes to the `slime-connection-list-mode-map` make sense. Have you considered the alternative of adding confirmation questions to the commands instead? What do you think about that alternative?

attila-lendvai:
latin-1 is not exotic, but if the text file contains sequences that are illegal in utf-8, then it's a sign that the file is probably not latin-1 encoded. but i'm happy with what you proposed, as long as if i can set the lisp default to utf-8.

re connection list: i wouldn't call confirmation questions an alternative, but rather an extra layer of safety, and they sound like a good idea.

luismbo:
Interpreting latin-1 as UTF-8 definitely leads to trouble:

```
CL-USER> (babel:octets-to-string (babel:string-to-octets "Luís" :encoding :iso-8859-1)
                                 :encoding :utf-8)
Illegal :UTF-8 character starting at position 2.
   [Condition of type BABEL-ENCODINGS:INVALID-UTF8-CONTINUATION-BYTE]
```

So, you wouldn't be happy with simply adding confirmation questions but not changing the bindings, right?

I'll finish the clean up branch tomorrow and put it up for review. Thanks Attila. :)

attila-lendvai:
nm, my knowledge of encodings seems to have a few holes. but the question is moot anyway if we switch to relying on the inferior lisp's default encoding, which i think is the solution with the least surprises.

confirmation questions may be enough, but when i first encountered those functions, their names also misguided me. i kept on restarting the web service from under live users when i thought i'm merely disconnecting slime...

i suggest a more complex key shortcut (at least a C-something), plus a question, plus renaming the functions. i can offer a patch for that once you're done with the cleanup branch.

luismbo:
https://github.com/luismbo/slime/tree/pr/107 has all your changes I believe. I backed down on two of the sldb-faces because I was afraid it was starting to look to much like a christmas tree. :) Let me know if you're happy with branch.

attila-lendvai:
i have no time to play with it now, but the patch looks perfect, thanks for your work Luis! looking forward to seeing it merged.

(the faces i can always customize for myself, i just wanted more visibility for newcomers that faces are used/defined at all.)

attila-lendvai:
i guess we can close this pull request now, and you can create a new one from your branch?

luismbo:
Pushed the branch onto master. Thanks Attila. :)

luismbo:
The slime-connection-list changes have been reverted by 83513fd. @ellerh what are your objections? Do you object to both the confirmation messages and the changed bindings?

ellerh:
On Wed, May 20 2015, Luís Oliveira wrote:

> The slime-connection-list changes have been reverted by 83513fd. @ellerh what
> are your objections? Do you object to both the confirmation messages and the
> changed bindings?

Both.

I often kill the subprocess and the confirmation message annoys me.
Even more often, I have to restart the subprocess.  If you want to play
nanny and be consistent then you should also ask for confirmation when
killing the process before restarting it.

I also doubt that hitting C-k or R by accident happens easily;
especially as one has to bring up the connection list buffer first.
Using C-c as prefix key is overkill when the buffer has it's own keymap
with only a few commands.

I rarely need the disconnect command, but why k would be a particularly
"reasonable" key for it, eludes me.  I would bind it to D.

There are few arguments why Mr. Lendvai's keys are better.  Changing key
bindings now would gain little (if anything) but annoys everybody who is
used to the traditional behavior.

Helmut

luismbo:
Fair enough, your arguments make sense to me. Sorry about the noise. What about adding a new disconnect command (bound to D) and renaming the current disconnect to kill-inferior-lisp (or whatever), but keeping the C-k binding. Does that seem reasonable?

I suppose the disconnect command is useful when connecting to long-lived images. Things like web servers, for instance.

ellerh:
On Wed, May 20 2015, Luís Oliveira wrote:

> Fair enough, your arguments make sense to me. Sorry about the noise. What
> about adding a new disconnect command (bound to D) and renaming the current
> disconnect to kill-inferior-lisp (or whatever), but keeping the C-k binding.
> Does that seem reasonable?

Yes, that sounds good.

attila-lendvai:
you guys should have made a proper fork, where the standards are not conservatism, but

attila-lendvai:
hehh, you guys should have made a proper fork, where the standards are not conservatism and commit wars, but rational debate.

binding disconnect to something 'd'-related is indeed much more reasonable than 'k', but 'd' was already used up for setting the default connection, and using shift for commands just felt alien to me. and i was probably still disturbed by the experience of repeatedly killing the server processes unintentionally from below the users and that probably pushed me to the other extreme with those shortcuts...

and that "Mr. Lendvai"! you know Helmut, i don't even know what else to react besides pointing that out... that's just beyond my words right now... :)

attila-lendvai:
> binding disconnect to something 'd'-related is indeed much more reasonable than 'k'

actually, the reason i chose 'k' is that in the archaisms of emacs you kill stuff: words, paragraphs, buffers, etc...

luismbo:
@attila-lendvai yeah, I see what you mean. But, that killing concept means something else doesn't it? It means "cut" in modern "cut/copy & paste" lingo.

This pull request is not forgotten, BTW. I just haven't had the time to implement the approach outlined in my previous comment. Feel free to send a patch / pull request implementing that, BTW. :-) That'd be very helpful. (Extra points if you do the dreaded ChangeLog dance, but a ChangeLog-less patch would be OK, too.)

attila-lendvai:
dunno. buffers are also killed, but then windows are deleted... i haven't found too much consistency.

but i also don't have a strong opinion about this, i just wanted to point out the reasoning behind my original decision, now that i recalled it.

luismbo:
Oh, you're right about buffers. :-) I guess Helmut's point is that we'd be binding "k" to a disconnect, not a kill command.

attila-lendvai:
@luismbo see the updated patch as discussed. it removes the yes-no questions and binds the functions as discussed in https://github.com/slime/slime/pull/107

(the other changes have all been merged since then)

attila-lendvai:
any words on this?

PR #172, opened 2014-07-08
Add the option to look for history from (mark).

kc-:
The behaviour is controlled through slime-repl-history-use-mark -- when
nil, the original history replacement is done.
- Add a `slime-repl-history-yank-start' function that returns the
  furthermost position from (mark) and the start of the current input.
- Replace input from slime-r-h-yank-start instead of the beggining.

For example:

> CL-USER> (foo (bar :baz 23))
> 42
> CL-USER> (let ((x **C-SPC** (foo **M-p**
>         ;; will replace from `(foo' to yield:
> CL-USER> (let ((x (foo (bar :baz 23))

### DIFF
--- a/contrib/slime-repl.el
+++ b/contrib/slime-repl.el
@@ -129,6 +129,23 @@ current repl's (as per slime-output-buffer) window."
 (defvar slime-repl-input-start-mark)
 (defvar slime-repl-prompt-start-mark)
 
+(defvar slime-repl-history-use-mark nil
+  "A non-nil value means that history will be replaced from the mark.
+
+Instead of replacing form input-start, look up history and replace input
+from the mark. Calling 'slime-repl-previous-input',
+ 'slime-repl-previous-matching-input' or their -next counterparts with a prefix
+ argument sets this variable for the duration of one history lookup.")
+
+(defun slime-repl-history-yank-start ()
+  "The position which 'slime-repl-previous-input' will replace from.
+
+When 'slime-repl-history-use-mark' is non-nil, and (mark) is after the current
+input start, return it.  Otherwise, return 'slime-repl-input-start-mark'."
+  (if (and slime-repl-history-use-mark (mark))
+      (max (mark) slime-repl-input-start-mark)
+      slime-repl-input-start-mark))
+
 (defun slime-output-buffer (&optional noprompt)
   "Return the output buffer, create it if necessary."
   (let ((buffer (slime-connection-output-buffer)))
@@ -614,7 +631,7 @@ The input is the region from after the last prompt to the end of
 buffer."
   (or (run-hook-with-args-until-success 'slime-repl-current-input-hooks
                                         until-point-p)
-      (buffer-substring-no-properties slime-repl-input-start-mark
+      (buffer-substring-no-properties (slime-repl-history-yank-start)
                                       (if until-point-p
                                           (point)
                                         (point-max)))))
@@ -811,7 +828,7 @@ earlier in the buffer."
 (defun slime-repl-delete-current-input ()
   "Delete all text from the prompt."
   (interactive)
-  (delete-region slime-repl-input-start-mark (point-max)))
+  (delete-region (slime-repl-history-yank-start) (point-max)))
 
 (defun slime-eval-last-expression-in-repl (prefix)
   "Evaluates last expression in the Slime REPL.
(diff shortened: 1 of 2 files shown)


kc-:
380aa298 has the changeset against the current master/head.

contrib/slime-repl.el
```diff
@@ -129,6 +129,19 @@ current repl's (as per slime-output-buffer) window."
...
 
+(defvar slime-repl-history-use-mark 'T
+  "When T, use the mark as starting point for history lookup and replacement.")
+
+(defun slime-repl-history-yank-start ()
+  "The position of the current input which
+  slime-repl-previous-input will replace from.  When
+  slime-repl-history-yank-start is T, and (mark) is after the
+  current input start, return it.  Otherwise, return
+  slime-repl-input-start-mark."
```
luismbo: Don't indent the docstring, use quotes around the functions you're referencing, and write T in downcase. (Or non-null, which sounds a bit too technical IMHO but is the idiomatic Emacs way.) See <https://www.gnu.org/software/emacs/manual/html_node/elisp/Documentation-Tips.html> for many more details.

contrib/slime-repl.el
```diff
@@ -717,6 +730,7 @@ With prefix argument send the input even if the parenthesis are not
 balanced."
   (interactive "P")
   (slime-check-connected)
+  (setf (mark) slime-repl-input-start-mark)
```
luismbo: What does this bit accomplish?

stassats:
If it's on by default, why even have an option? Can it break somebody's workflow?

contrib/slime-repl.el
kc-: You're absolutely right, this is superfluous.

kc-:
> If it's on by default, why even have an option? Can it break somebody's workflow?

It's possible: if somebody were to use kill-region for some partial input at the repl, then intended to scroll through the history, the replacement would not happen from where that someone expects.

I've changed it to be off by default.

luismbo:
@kc- 🤔 as an alternative to an option, have you considered using distinct commands and keybindings? For instance, <kbd>M-R</kbd> would bind to a new `slime-repl-backward-input-from-mark` command, or <kbd>C-u M-r</kbd> could pass an optional argument to `slime-repl-backward-input` that would make it search from the `(mark)`.

This would have the advantage of being one fewer option (options are bad, they make features less accessible and make them harder to test, etc). I'm not sure how cumbersome the alternative keybindings are though, but heavy users of this feature could then rebind `M-r` to suit their preference.

kc-:
No, I hadn't considered that. There are no strong feelings one way or the other on my part, I followed somebody's lead when we talked about it some years ago. Here's the logged conversation: https://irclog.tymoon.eu/freenode/%23lisp?around=1402034564#1402034564

I understand the notion about options obscuring features, so which alternative do you prefer? Additional function or prefix argument?

luismbo:
@kc- I'm currently leaning towards the <kbd>C-u M-r</kbd>; I think this approach will make this feature more easily discoverable by users and is not intrusive. (The example you brought up could indeed be very confusing, particularly since the mark is invisible.) But I don't have particularly strong feelings either. If you agree and go with that route, I suggest you document this new <kbd>C-u</kbd> behaviour in `slime-repl-backward-input`'s docstring; likewise for `slime-repl-forward-input`.

stassats:
> > If it's on by default, why even have an option? Can it break somebody's workflow?
> 
> It's possible: if somebody were to use kill-region for some partial input at the repl, then intended to scroll through the history, the replacement would not happen from where that someone expects.
Shouldn't M-p do the optional thing only when there's a selected region? How would killing a region affect that behavior?

luismbo:
@stassats no, this proposed feature does not use regions at all.

luismbo:
@kc- what about this alternative approach to the use case of yanking a previous expression into the current expression? We could have a variant of <kbd>M-r</kbd> (`slime-repl-previous-matching-input`) that insert the matching input at point instead of replacing the current expression. (It could be implemented as an optional argument, i.e. <kbd>C-u M-r</kbd>.)

kc-:
After some thought, I like `C-u M-p` (/ `M-p`) as well -- thank you for both suggestions, by the way -- I'll push the changes when they're ready. The part about matching only from `(mark)` should stay, in my opinion; then, if `input-start` is past `(mark)`, we use an empty prefix and match everything, but insert at the point like you describe. Would you agree?

kc-:
> > > If it's on by default, why even have an option? Can it break somebody's workflow?
> > 
> > 
> > It's possible: if somebody were to use kill-region for some partial input at the repl, then intended to scroll through the history, the replacement would not happen from where that someone expects.
> 
> Shouldn't M-p do the optional thing only when there's a selected region? How would killing a region affect that behavior?

In the original proposal: yes and no, I'm not confident enough in my emacs terminology: If mark == point, the region is empty, so there isn't any selection, correct? In that case the answer is no, the original proposal does not only apply to a selected region, because it would insert history elements at the point while keeping some typed prefix untouched -- with an empty search prefix, every history item is a match.

kc-:
@luismbo e70309d contains an implementation in the way I think we both agree it should be done, with the exception that I did not include the case of inserting history at `point` without setting `mark` first.
The reason is that expanding history items moves the point to the end of the line, and I didn't want to add the additional complexity of memorizing/resetting the insertion point.

There's still `slime-repl-history-use-mark`, which is nil by default but lexically bound to the prefix-arg in the -forward and -backward browsing functions, for both `M-p` and `M-r`. This way, users that want the functionality can simply set the var to t permanently.

contrib/slime-repl.el
```diff
@@ -1017,15 +1034,21 @@ If EXCLUDE-STRING is specified then it's excluded from the search."
   "Cycle backwards through input history.
 If the `last-command' was a history navigation command use the
 same search pattern for this command.
-Otherwise use the current input as search pattern."
+Otherwise use the current input as search pattern.
+
+With a prefix-arg, do replacement from the mark."
   (interactive)
-  (slime-repl-history-replace 'backward (slime-repl-history-pattern t)))
+  (let ((slime-repl-history-use-mark current-prefix-arg))
```
luismbo: Since `slime-repl-history-use-mark` is unconditionally bound here, setting it to `t` has no effect, right?

kc-: Right 🤦‍♂️ thanks for catching that.

Are you happy with the behaviour outside of that fault?

luismbo: Yes, WFM.

luismbo:
Thanks @kc-.

PR #231, opened 2015-02-19
Restore the mdot-fu contrib

### LINKED ISSUE #111: slime-mdot-fu contrib is broken
http://lists.common-lisp.net/pipermail/slime-devel/2010-August/017676.html

csziacobus:
More precisely, the enclosing context contrib. A clean-up commit
previously broke the code, but some functions that were deleted are
re-introduced in order to make local meta points work again. It'd
be nice in the future to clean this up. Fixes #111

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,10 @@
+2015-02-19  Charles Zhang  <carolusiacobus@gmail.com>
+
+	* contrib/slime-enclosing-context.el: Restore some long dead functions
+	in order to make the slime-mdot-fu contrib work again.
+	slime-edit-local-definition and meta-point on local
+	variable and function definitions work as a result.
+
 2015-02-12  Stas Boukarev  <stassats@gmail.com>
 
 	* swank/sbcl.lisp (definition-source-buffer-and-file-location):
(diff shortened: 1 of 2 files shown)


csziacobus:
Rebased against HEAD

luismbo:
@csziacobus what was the commit that broke it? Also, could you expand on what needs cleaning up in your opinion?

csziacobus:
It was commit 67b8bb2ecd8e7fcbf4d96783ac578de040cacf98, which deleted many many functions a couple years ago after a rewrite, afaik (i dunno, i only took a cursory glance at the history). Some of the functions that enclosing-context depended on were removed, and all I really did in order to restore the contrib was restore the 3 or 4 functions deleted that were needed to work. I mention cleaning up the contrib because the code I restored was very general and part of an old way of doing parsing; therefore it'd be better to rewrite the enclosing-context contrib to not use that outdated parsing functionality that was superseded. However, doing that is not really important as local meta-point works now without any problems and only a very small subset of that old parsing functionality was reintroduced, so not a big deal.

luismbo:
@csziacobus cool, thanks for your clarifications. Pushed your commit with minor tweaks: moved your ChangeLog file to the right file (if you used <kbd>C-x 4 a</kbd> Emacs would have helped you there) and removed some trailing whitespace because OCD. :-) Thanks!

PR #232, opened 2015-02-21
Add documentation for mdot-fu and add to fancy contrib

csziacobus:


luismbo:
Since mdot-fu has been disabled for so long, I'm inclined to be conservative and only enable it by default after a month or two of daily testing. Do you agree?

csziacobus:
Sure, that's more than reasonable. I guess we can just merge the documentation then so people will know it exists and can test.

luismbo:
Oh, sure. Care to rewrite the patch to keep the documentation bits only?

doc/slime.texi
```diff
@@ -3177,6 +3178,11 @@ Toggle exclusion of swank functions from the report.
 
 @end table
 
+@node SLIME Enhanced M-.
+@section SLIME Enhanced M-.
+@code{slime-mdot-fu} enables meta-point to jump to local variables bound with @code{let} and @code{let*},
```
luismbo: I think you should `M-x fill-paragraph` this. (And I think slime.texi assumes `fill-column` = 70.)

csziacobus:
Commit amended.

luismbo:
Commited via 72701d5. I noticed a couple of days ago that, unlike what you (and I) expected, slime-mdot-fu works for `LET` but not `LET*`. I believe I've fixed it in 49673e9. Thanks!

PR #233, opened 2015-02-22
Suggest downcase completions for downcase users by default.

orivej:
This change makes SLIME suggest completions after `package:` in lower case when `*print-case*` is `:downcase`. The idea is that if the user has chosen printing symbols in lower case, he does not want to read them in upper case in the completion window.

Upcase users may also want lower case completions by default, and this can be handled by a new special variable in the `swank` package. If so, I can extend or provide another patch and document the new variable, but I think that `*print-case*`-based logic should be there, for the new variable may not be customized (unbound).

### DIFF
--- a/contrib/swank-c-p-c.lisp
+++ b/contrib/swank-c-p-c.lisp
@@ -163,7 +163,9 @@ Return these values:
 INPUT is used to guess the preferred case."
   (ecase (readtable-case *readtable*)
     (:upcase (cond ((or with-escaping-p
-                        (not (some #'lower-case-p input)))
+                        (if (eq *print-case* :downcase)
+                            (some #'upper-case-p input)
+                            (not (some #'lower-case-p input))))
                     #'identity)
                    (t #'string-downcase)))
     (:invert (lambda (output)


orivej:
What prevents this from being merged?

luismbo:
@orivej Pardon my ignorance, but I can see the effect of your patch? I already get lowercase completions regardless of `*print-case*`. What am I missing?

orivej:
Type `cl:` in Slime REPL and press tab. `*Completions*` buffer then begins with:
```
Click on a completion to select it.
In this buffer, type RET to select the completion near point.

Possible completions are:
cl:&ALLOW-OTHER-KEYS 	cl:&AUX
cl:&BODY 	cl:&ENVIRONMENT
```
with `&ALLOW-OTHER-KEYS` and other symbols in upper casee.

luismbo:
Hmm, I get those in lower case. (Using a vanilla setup I get the same output as you do, though.) I'll try to figure out what I am doing to get that and get back to you.

luismbo:
`(setq slime-complete-symbol-function 'slime-fuzzy-complete-symbol)` is it. If I use that once, I always get lowercase completions afterwards. Can you figure out why?

luismbo:
Oh, nevermind that. I was assuming `slime-simple-complete-symbol` was the default value. And that also downcases symbols. Ahrg, I'm too sleepy. I'll try this again tomorrow. Sorry for the noise.

orivej:
Sorry, this patch is for `slime-c-p-c` (loaded with `slime-fancy`), so I forgot to mention earlier that I assume `(setq slime-complete-symbol-function 'slime-complete-symbol*)` (performed on load of `slime-c-p-c.el`).

luismbo:
So, all the other symbol completion methods have the downcase behavior you (and I) want. None of them pay any attention to `*print-case*` because the idea is that the completed symbol should yield the correct symbol when read back, and that is related to `*readtable-case*`.

The three methods should probably share this logic. (If you're inclined to go for that refactoring, definitely give it a go! :-)) Meanwhile, this seems to make sense:

```
diff --git a/contrib/swank-c-p-c.lisp b/contrib/swank-c-p-c.lisp
index 38aae3b..0383788 100644
--- a/contrib/swank-c-p-c.lisp
+++ b/contrib/swank-c-p-c.lisp
@@ -163,7 +163,7 @@ Return these values:
 INPUT is used to guess the preferred case."
   (ecase (readtable-case *readtable*)
     (:upcase (cond ((or with-escaping-p
-                        (not (some #'lower-case-p input)))
+                        (some #'upper-case-p input))
                     #'identity)
                    (t #'string-downcase)))
     (:invert (lambda (output)
@@ -173,7 +173,7 @@ INPUT is used to guess the preferred case."
                        (upper (string-downcase output))
                        (t output)))))
     (:downcase (cond ((or with-escaping-p
-                          (not (some #'upper-case-p input)))
+                          (some #'lower-case-p input))
                       #'identity)
                      (t #'string-upcase)))
     (:preserve #'identity)))
```

Do you agree with this approach? Does it work for you?

luismbo:
pinging @orivej :)

orivej:
Sorry for the delay; yes this works for me and I'm happy with such resolution.
> The three methods should probably share this logic.
From the quick glance, c-p-c and fuzzy actually do. (The latter calls `completion-output-symbol-converter` from the former, but with a package name being part of the argument, so that its case is taken into account.) Simple completion seems to ignore input case, but I have not tried to use it.

luismbo:
Cool, thanks for your feedback @orivej!

PR #234, opened 2015-02-24
Fix lexical binding headers.

Bad-ptr:
Added the `-*- lexical-binding: t; -*-` to the first line everywhere where cl-lib is used(mostly in contrib/ directory).
Also replaced some of the `lexical-let`s with `let`s(cause it must be the same when lexical-binding is t) and so removed cl requirement.
Looks like it didn't break anything(though i didn't test everything, to be honest).


luismbo:
The 70s called, they confirm that lexical binding is indeed the future! ;-) That being said, last time we had this discussion, we weren't ready to drop Emacs 23. @ellerh, any news on that front?

luismbo:
Oddly enough, all tests pass on Emacs 23, albeit with lots of warnings.

Bad-ptr:
I think that only b7c35bd e9a18b2 4ec0f6f 0396c00 commits could be incompatible with old emacses.

ellerh:
>  That being said, last time we had this discussion, we weren't ready to drop Emacs 23. @ellerh, any news on that front?

My policy is to support at least the Emacs version that is in Debian Stable.  Currently that's Emacs 23.  When Debian Jessie comes out, maybe in spring, then it's Emacs 24.

Bad-ptr:
@ellerh 
> Emacs 23
Ok, i removed all chages that could break emacs 23.
Also in `contrib/slime-presentations.el` the `cl-labels` macro is used(I don't touch anything it already was there), it will not work in old emacs(in which lexical-binding is not supported).

attila-lendvai:
FTR, current debian stable has emacs 24 now.

luismbo:
Whoo hoo. @Bad-ptr do you want/need to revise your patch to embrace Emacs 24-ness?

PR #235, opened 2015-03-11
Swank-presentation-streams needs to update is wrapper for sb-impl::%print-unreadable-object.

bobbysmith007:
The body argument is optional now in sb-impl::%print-unreadable-object and calls to this were erroring in new SBCL(1.2.9).


jimka2001:
I encountered this problem as well.  A similar fix solved the problem for me.

luismbo:
Pushed as 0311197 with minor tweaks, thanks!

PR #240, opened 2015-03-31
ABCL fixes for inspecting local arguments

easye:
Current code has "off by one errors".  This fixes that as well as attempting to match generalize lamba lists to the available compile-time names.

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2015-03-31  Mark Evenson  <evenson.not.org@gmail.com>
+
+	* swank/abcl.lisp (frame-locals): Triage for inspecting locals of
+	the stack.  Functionally better, partly able to resolve all lambda
+	lists by at least returning warnings rather than errors.
+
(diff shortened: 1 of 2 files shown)


luismbo:
If you have the energy, squash the two commits together and push them. (I think using keyword symbols for the loop directives is not the convention followed in the rest of SLIME, but that's fairly minor nitpicking.)

easye:
@luismbo Moved to single commit as requested in https://github.com/slime/slime/pull/241

easye:
On 3/31/15 15:34, Luís Oliveira wrote:
> If you have the energy, squash the two commits together and push them. (I think using keyword symbols for the loop directives is not the convention followed in the rest of SLIME, but that's fairly minor nitpicking.)
> 
> ---
> Reply to this email directly or view it on GitHub:
> https://github.com/slime/slime/pull/240#issuecomment-88087134
> 

Done as <https://github.com/slime/slime/pull/241/files>.

See you in London?

-- 
"No, this is not a disentanglement from, but a progressive /knotting-into/

PR #241, opened 2015-04-03
ABCL: swank/abcl.lisp (frame-locals frame-var-value)

easye:
@luismbo Squashing patches as requested.  

Fix off by one errors in inspecting arguments across stack frames.  Attempt to match
arguments to parameters names as recorded at compile time.


luismbo:
Pushed as 76da415. Thanks! (Sorry for taking so long.)

PR #244, opened 2015-05-06
quick hack to fix compile warnings on allegro

rick-monster:
slime-compile-and-load-file was crapping out when I hit not-yet-defined functions under allegro/windoze so here's a quick and dirty fix...

Works on my machine!

### DIFF
--- a/swank/allegro.lisp
+++ b/swank/allegro.lisp
@@ -425,6 +425,8 @@
 
 ;; TODO: report it as a bug to Franz that the condition's plist
 ;; slot contains (:loc nil).
+;; HACK: to get round this broken behaviour fall back on end for
+;; location of compile error
 (defun handle-undefined-functions-warning (condition)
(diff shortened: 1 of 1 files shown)


luismbo:
What do you mean by "crapping out"?

rick-monster:
'crapping out' in this case means that the call to (1+ pos) throws you into the debugger (because the variable start gets bound to nil in that destructuring bind)

luismbo:
Can't seem to reproduce it. Can you give me a recipe?

rick-monster:
hmm let me see if I can reproduce in a bare test case.  Encountered the bug whilst refactoring a pretty substantial codebase...

rick-monster:
ok so apologies for posting this here - the bug in slime-compile-and-load is caused by some interference from reader macros in aforementioned codebase.

Sorry once more to waste your time!

luismbo:
@rick-monster no worries. :)

PR #253, opened 2015-06-30
replaced call to values with cl-values.  There is no such function as values

jimka2001:
This PR fixes broken slime-edit-definition

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,3 +1,9 @@
+2015-06-30 Jim Newton <jimka.issy@gmail.com>
+
+	* contrib/slime-enclosing-context.el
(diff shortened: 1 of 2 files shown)


jimka2001:
A little more description of the problem this fix solves.

Before this fix, if I pressed M-. to jump to the definition point, I would get the message in the mini buffer:
   Symbol's function definition is void: values

luismbo:
Thanks, pushed 0f212b3 with minor tweaks. (In particular the ChangeLog syntax was slightly off. It's a minor detail, sure, but it's easily avoided if you use `M-x add-change-log-entry-other-window` which is bound to `C-x 4 a`.)

PR #257, opened 2015-07-08
* slime-media.el: Base64-decode literal image data

ecraven:
If image data is passed as :data, assume it is base64-encoded and decode
it. If there is no :data, don't change anything.

This fixes problems with sending actual binary data over the wire.

### DIFF
--- a/contrib/slime-media.el
+++ b/contrib/slime-media.el
@@ -10,11 +10,17 @@
   (:on-load
    (add-hook 'slime-event-hooks 'slime-dispatch-media-event)))
 
+(defun slime-media-decode-image (image)
+  (mapcar (lambda (image)
+	    (if (plist-get image :data)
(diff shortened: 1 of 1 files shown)


ecraven:
I've played around with sending images, and ran into problems with binary data. Sending images as :file works fine, but :data broke often. Base64-encoding the data (and decoding it in emacs) seems to be a simple solution.

From what I've found on the internet, no-one except slimer is using slime-media, and they only seem to use :file images, not :data.

ecraven:
slime-media

csrhodes:
Looks sane to me... anyone else with an opinion?

luismbo:
Pushed with minor tweaks as 76632da. Thanks!

ecraven:
thank you very much for the swift merging!

PR #259, opened 2015-07-09
using (in-readtable) to select a readtable in a source file buffer

budden:
I also found a problem: I can't change readtable through (in-readtable) in a slime repl buffer. I guess *readtable* somehow gets bound around real evaluation, but I dont know how to avoid it. Currently I have no idea what to do. It looks like some special actions must be done for repl buffers as we must take *readtable* 

So I created a pull request, but code evidently needs fixing. 

Sorry code is not written according to all standards (yet).


budden:
Now I see this "problem" is just a feature. Package cl-user is bound to some readtable in swank:*readtable-alist* . So regardless of global value of *readtable*, swank binds *readtable* to the value from *readtable-alist*. 

When I do
cl-user> (defpackage :foo (:use :cl))
cl-user> (in-package :foo)
foo> (assoc "FOO" swank::*readtable-alist*)
nil
foo> (in-readtable :buddens-readtable-a)
everything is ok - readtable changes.

budden:
I also found that the following definition from named-readtables adds complexity: 

(defmacro in-readtable (name)
  ....
  (%frob-swank-readtable-alist *package* *readtable*)...)

where 
(defun %frob-swank-readtable-alist (package readtable)
 ... 
 (pushnew (cons package-name rt) *readtable-alist*)
...

so, at every invocation of (in-readtable), 
package and readtable stick to each other and we can't keep them truly orthogonal. 
Some other form is needed to specify readtable for file, not in-readtable.

luismbo:
Perhaps the presence of named-readtables and in-readatable should override swank's readable alist. Does that sound like a good idea?

budden:
Hi! 

I don't understand what do you mean by "this" now. If you mean behaviour of (in-readtable) which implicitly changes readtable-alist, I think that this is not a good idea. But if it is used by many people, this is not a bad idea any more, but just a Tradition. Most of the time I use only two readtables - "standard" and "mine" which are mostly compatible to each other. And I didn't notice the problem before. Workaround is not to use "in-readtable", but to add another special form with no side effects but changing *readtable*. Say (in-readtable-2). Things get bit comlicated this way, but tolerable. Third variant (IMO best in the long term) is to alter in-readtable not to change *readtable-alist*, and to introduce accessor package-readtable for changing *readtable-alist* but I think there will be strong resistance of users. 

If you mean my proposal as a whole, I feel it can break something. I don't know. Maybe current infrastructure is sufficient already. On the other hand, CL generally tries to keep things orthogonal. My third variant goes this way, while in-readtable does not.

luismbo:
Nevermind, I shouldn't try to discuss technical matters late at night. :)

Anyway, you mentioned that you wanted to clean up the code. I suggest you go ahead and do that before I review it.

budden:
No, I wanted to clean up my brain and I succeeded in the process :)
I see no bugs, but a design problem, which I explained in previous post.

PR #260, opened 2015-07-13
Fixed autoload spec of ‘hyperspec-lookup’

mrkkrp:
Previously name of file to load was incorrect.


mrkkrp:
Any update on this PR? I would like to use `hyperspec-lookup`.

luismbo:
Can you add a ChangeLog entry?

mrkkrp:
@luismbo, here it is.

mrkkrp:
Diff is weird, I pulled all you've done while my PR hanged here, and then force pushed to be even with your `master`. Now my local `master` is OK, but the diff here is from the first try (that caused merge conflicts, of course). I better open new pull request.

PR #261, opened 2015-07-18
Fix autoload spec of the ‘hyperspec-lookup’ function

mrkkrp:


mrkkrp:
@luismbo, change log is here.

ellerh:
A directory ("lib/") in a autoload specification is a bit unusual.  But I suppose we will see if it causes problems.

mrkkrp:
@ellerh, Usual or not it just reflects how SLIME code is currently organized. With this fix I can call `hyperspec-lookup` anytime I want and it won't say that `"hyperspec"` file doesn't exist :-)

PR #265, opened 2015-08-25
Update README.md

chuchana:


luismbo:
I'm not 100% sure about this out-of-nowhere reference to use-package. Hmm, what about an entry in the manual with a complete use-package snippet? (We can link to that from the README.md.)

chuchana:
Yes, I was not really sure either. I had just spent a bit of time figuring out why the REPL did not show up with use-package because I had overlooked the "before" in the text. I thought emphasising it was probably not necessary for normal use, so I added the use-package specific part because I wanted to do something that might help someone with the same problem. If you think I should add a use-package snippet to the manual, I can do that, but other than whether to put things in the init block or the config block, it is pretty straightforward. If you think it should just stay the way it is, just close this.

luismbo:
I think being able to copy & paste a use-package snippet is useful.

chuchana:
Ok. I have created a new PR under #268 because I did not know how to add the changes here.

PR #266, opened 2015-09-01
swank-presentation-streams/slime-stream-p fixes

zellerin:
Hi,

looking at the sources swank-presentation-streams, I spotted some strangeness in `slime-stream-p`:
- It caches return value for `t` stream designator,
- It does not follow clhs convention for what `t` and `nil` stream designators mean.

On the other hand, I do not see whether this function is ever called with stream designator that is not a stream (`print-object` calls that indirectly, and it is specified in clhs to accept stream; so the only option is someone calling presenting-object directly with t or nil).

On the third hand. the caching could also fail on synonym streams. Does it make sense to resolve them as well?

Best regards,
Tomas

### DIFF
--- a/contrib/swank-presentation-streams.lisp
+++ b/contrib/swank-presentation-streams.lisp
@@ -89,8 +89,9 @@ Two special return values:
 :DEDICATED -- Output ends up on a dedicated output stream
 :REPL-RESULT -- Output ends up on the :repl-results target.
 "
-    (if (eq stream t) 
-	(setq stream *standard-output*))
+    (case stream
+      ((nil) (setq stream *standard-output*))
+      ((t)  (setq stream *terminal-io*)))
 
     (if (eq last-stream stream)
 	last-answer
--- a/contrib/swank-presentation-streams.lisp
+++ b/contrib/swank-presentation-streams.lisp
@@ -89,12 +89,13 @@ Two special return values:
 :DEDICATED -- Output ends up on a dedicated output stream
 :REPL-RESULT -- Output ends up on the :repl-results target.
 "
+    (if (eq stream t) 
+	(setq stream *standard-output*))
+
     (if (eq last-stream stream)
 	last-answer
 	(progn
 	  (setq last-stream stream)
-	  (if (eq stream t) 
-	      (setq stream *standard-output*))
 	  (setq last-answer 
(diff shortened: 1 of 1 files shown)


luismbo:
I suppose slime-stream-p and the caching mechanism expect an actual stream and not stream designators. Perhaps we can add a `check-type` to make that explicit? What do you think about that alternative?

zellerin:
I guess it depends on what the plans with presentation code are. If it is supposed to be kept internal and used only from inside of the rest of swank, your option is clean and definitely good. It was just the special casing of `t` what I did not like.

If you plan to expose presentations interface outside of the swank, then explicitly allowing designators would be more friendly, but still not necessary.

As a side note, I consider allowing program writers to make an interface for their program in emacs using `presenting-object` and extending `menu-choices-for-presentation` as possibly nice feature, even though the symbols are currently not exported, the option is not much documented/advertised and I find a bit tricky to get hands on slime streams. Would a patch to export the symbols and document `presenting-object` (this time with changelog-like commit messages, I got a bit late to reading your guidelines) have a chance to be accepted?

luismbo:
Having an interface for augmenting the possibilities of interacting with objects via presentations is definitely welcome IMO!

swank-presentation-streams is something different though, isn't it? IIUC, it's a mechanism that hooks into regular streams and changes the output to inject presentation markers whenever an object is being written. It's a very neat idea, I think, but it has been historically problematic, at least on SBCL, mostly because it messes with (redefines!) implementation details.

Please, do share your plans, they sound interesting. Why do you plan to export `slime-stream-p`?

zellerin:
Plans: I put on https://github.com/zellerin/slime-presentations-example/blob/master/adventure.lisp a commented example of what I am doing (technically; I am not in fact working on an adventure game, and it might be bad use case for several reasons anyway; what I really do is explore and act on a connected data set) and in the first package I (re)exported  what I need as an interface (it is not much, and it does *not* contain `slime-stream-p`). 

The only relevance to current patches is that they allow to use `nil` and `t` as parameters for the `presenting-object` macro. Thinking about it, I might want to wrap with some utility layer anyway, so this is actually not such an issue anyway. I have no reasons to export `slime-stream-p`, I just need `presenting-object` to be predictable with regard to the stream parameter.

PR #270, opened 2015-10-03
Draft / PoC: bind swank to a unix socket

adlai:
The functionality works (on CCL), but this PR is not intended to be merged as-is; it's more of an RFC. If this is useful to you, please at least "ACK"; better yet, submit code for missing backends and/or fix [the `socket-quest` restart](https://github.com/slime/slime/blob/2cca7e/swank.lisp#L789-L793) to support non-integer "ports".

### DIFF
--- a/swank/ccl.lisp
+++ b/swank/ccl.lisp
@@ -102,12 +102,13 @@
   :spawn)
 
 (defimplementation create-socket (host port &key backlog)
-  (ccl:make-socket :connect :passive :local-port port
-                   :local-host host :reuse-address t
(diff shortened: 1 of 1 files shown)


luismbo:
I'm curious, what are the use cases for using unix sockets?

adlai:
Better access granularity: they require login access and file permission, rather than just port visibility.

adlai:
For reduction of pending requests, I close this one unmerged.

adlai (commented):
it was nice and quiet and useful for me

PR #277, opened 2015-11-04
add significant whitespace to summary line

tarsius:


luismbo:
Thanks, merged. Was MELPA (or package.el) having trouble with it or was it just a style change?

tarsius:
`lm-summary` from the built-in library `lisp-mnt` had problems. Melpa seems to be using more forgiving code.

PR #280, opened 2015-11-21
Protect updating indentation of deleted package.

MatthewRock:
Swank did not perform check whether the package whose symbols it will access exists. This resulted in errors in some cases. This patch fixes this issue.

For example of code that causes the issue, please refer to [my sandbox](https://github.com/MatthewRock/cl-sandbox) package; if you run `(sandbox:start)` and then `(sandbox:exit)`, you should get an error(depending on your setup you might get it every time, or only at the first time).

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -3590,24 +3590,25 @@ If FORCE is true then start again without considering the old cache."
   "Update the cache and return the changes in a (SYMBOL INDENT PACKAGES) list.
 If FORCE is true then check all symbols, otherwise only check symbols
 belonging to PACKAGE."
-  (let ((alist '()))
-    (flet ((consider (symbol)
-             (let ((indent (symbol-indentation symbol)))
-               (when indent
-                 (unless (equal (gethash symbol cache) indent)
-                   (setf (gethash symbol cache) indent)
-                   (let ((pkgs (mapcar #'package-name 
-                                       (symbol-packages symbol)))
-                         (name (string-downcase symbol)))
-                     (push (list name indent pkgs) alist)))))))
-      (cond (force
-             (do-all-symbols (symbol)
-               (consider symbol)))
-            (t
-             (do-symbols (symbol package)
-               (when (eq (symbol-package symbol) package)
-                 (consider symbol)))))
-      alist)))
+  (when (package-name package)
+    (let ((alist '()))
+      (flet ((consider (symbol)
+               (let ((indent (symbol-indentation symbol)))
+                 (when indent
+                   (unless (equal (gethash symbol cache) indent)
+                     (setf (gethash symbol cache) indent)
+                     (let ((pkgs (mapcar #'package-name
+                                         (symbol-packages symbol)))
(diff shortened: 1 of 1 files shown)


MatthewRock:
Unfortunately, tests failed. Could someone please help me with understanding why they failed, so I can try to improve my fix attempt?

luismbo:
Having a look.

luismbo:
@MatthewRock sometimes `swank::update-indentation/delta-for-emacs` with a null package (namely when the force argument is true) and `(package-name nil)` yields an error. One way to reproduce this is to invoke `M-x slime-update-indentation`.

Do you have a standalone test case that doesn't involve `cl-sandbox`? Alternatively, can you show me the stack trace of the error you get with deleted packages?

MatthewRock:
Not sure, but you could try going with (progn 
        (defvar *x* (make-package :testpack :use '(:cl)))
        (setf cl:*package* *x*)
        (setf cl:*package* (find-package :cl-user))
        (delete-package *x*)). Generally, if you delete package and exit it afterwards in one function call, it should result in the bug. I will provide stack trace later.

luismbo:
Thanks for the example. If I evaluate it in two parts, as you suspected, I can reproduce the issue. I think the proper fix is this:

```diff
diff --git a/swank.lisp b/swank.lisp
index 7ee4fb7..e3f8b2e 100644
--- a/swank.lisp
+++ b/swank.lisp
@@ -3603,7 +3603,7 @@ belonging to PACKAGE."
       (cond (force
              (do-all-symbols (symbol)
                (consider symbol)))
-            (t
+            ((package-name package)
              (do-symbols (symbol package)
                (when (eq (symbol-package symbol) package)
                  (consider symbol)))))
```

What do you think?

MatthewRock:
Looks good to me, if it passes the tests then it should be ok. I don't know how force and checking for all symbols work, so it still might cause the bug, but it looks really unlikely, if we take into the account that I appear to be one of the first who experienced the bug, and it is caused by doing something unpopular; plus, it worked previously. Thanks, luismbo!

I have other question though - the file has quite interesting whitespace range - tabs, spaces, ^L, etc. My emacs(I use Prelude) has a hook that is run before save, which removes trailing whitespace(and maybe even normalizes the existing one - of that I am not sure), which would render the commit unreadable, but on the other hand - would improve the quality of the file. Is there some way that I could refactor whitespace and commit it?

MatthewRock:
Looks like errors are not caused by me, or am I wrong?

luismbo:
@MatthewRock it was some random connectivity failure, yes, retrying the build, fixed it.

luismbo:
Can you add a ChangeLog entry and rebase everything onto a single commit?

We should remove trailing whitespace from all of SLIME, as it messes with diffs and whatnot. WILLFIX. :)

luismbo:
@MatthewRock hmm, I have an indentation cache fix to send as well, so I'll just send your fix as well. Don't worry. Thanks! ;-)

PR #282, opened 2015-11-29
Update to swank-asdf.

fare:


capitaomorte:
Hi @fare, I obsviously have nothing against this, but @luismbo, watch out for allegro 8.1 users at you-know-where that might be affected by this change to a mandatory contrib, there might be a couple yet.

luismbo:
@fare, I think the bits in swank-loader that decide whether to load swank-asdf need to be updated to reflect the new requirement.

attila-lendvai:
on this note, i often had headaches when swank was loaded with swank-loader, and then :depends-on (:swank) didn't work because swank.asd was not found.

to fix that i've added this: https://github.com/attila-lendvai/slime/commit/623ea28278874e5916a49e056d922e6737cead62

and i'm wondering how to move on from here.

shall Swank just declare that ASDF is needed to load it? if so, then this patch will not be needed. or if swank-loader will remain, then i think this patch is needed (but i'm not sure about the pushnew, maybe that should be an append at the end of the search functions).

any thoughts?

fare:
As far as I'm concerned, I'd be quite glad to declare that SLIME requires ASDF 3, especially since all active implementations provide it, and ASDF 3 comes with a script to install itself on old implementations. I once posted a patch to enable just that, and could dig it out. That could notably vastly simplify swank-loader. But only the SLIME maintainers can decide.

attila-lendvai:
FTR, looks like ASDF head requires this slime patch.

fare:
Does 3.1.6.5 still need my patch? I hopefully reverted the part that required it.

But this patch is still a good idea.

And yes, if you're willing to load using asdf, I can send a much simplified swank-loader.

luismbo:
@fare if you want to go ahead with this patch regardless of the discussion about further asdfizing swank-loader, then please address my comment about swank-loader needing to not load swank-asdf if ASDF is not recent enough, because signalling an error (thus failing to load SWANK altogether) when ASDF is old is plain obnoxious.

attila-lendvai:
with ASDF head, slime HEAD and without this patch:

```
debugger invoked on a UNDEFINED-FUNCTION in thread #<THREAD "main thread" RUNNING {1007044723}>: The function ASDF/INTERFACE::LOAD-SYSDEF is undefined.

Type HELP for debugger help, or (SB-EXT:EXIT) to exit from SBCL.

restarts (invokable by number or by possibly-abbreviated name):
  0: [ABORT] Exit debugger, returning to top level.

(FDEFINITION ASDF/INTERFACE::LOAD-SYSDEF)
0] back

Backtrace for: #<SB-THREAD:THREAD "main thread" RUNNING {1007044723}>
0: (FDEFINITION ASDF/INTERFACE::LOAD-SYSDEF)
1: (SB-FASL::LOAD-FASL-GROUP #S(SB-FASL::FASL-INPUT :STREAM #<SB-SYS:FD-STREAM for "file /home/alendvai/.slime/fasl/2015-11-23/sbcl-1.3.1.6.hu.dwim.4-ce203fd-linux-x86-64/contrib/swank-asdf.fasl" {10070C52A3}> :TABLE #(353 SET *PACKAGE* "SWANK" #<PACKAGE "SB-IMPL"> SB-IMPL::%DEFVAR #<PACKAGE "SWANK"> *ASDF-PATH* #1="/home/alendvai/common-lisp/slime/contrib/swank-asdf.lisp" #2=#<SB-KERNEL:LAYOUT for SB-C:DEFINITION-SOURCE-LOCATION {100002D4D3}> #S(SB-C:DEFINITION-SOURCE-LOCATION :NAMESTRING #1# :TOPLEVEL-FORM-NUMBER 1 :FORM-NUMBER 2 :PLIST NIL) BOUNDP ...) :STACK #(2 SB-KERNEL:%SET-FDEFINITION LOAD-SYSDEF FDEFINITION ASDF/INTERFACE::LOAD-SYSDEF 0 8 #1# #2# NIL NIL NIL ...) :DEPRECATED-STUFF NIL :SKIP-UNTIL NIL) NIL)
2: (SB-FASL::LOAD-AS-FASL #<SB-SYS:FD-STREAM for "file /home/alendvai/.slime/fasl/2015-11-23/sbcl-1.3.1.6.hu.dwim.4-ce203fd-linux-x86-64/contrib/swank-asdf.fasl" {10070C52A3}> NIL NIL)
3: ((FLET SB-FASL::LOAD-STREAM :IN LOAD) #<SB-SYS:FD-STREAM for "file /home/alendvai/.slime/fasl/2015-11-23/sbcl-1.3.1.6.hu.dwim.4-ce203fd-linux-x86-64/contrib/swank-asdf.fasl" {10070C52A3}> T)
4: (LOAD #P"/home/alendvai/.slime/fasl/2015-11-23/sbcl-1.3.1.6.hu.dwim.4-ce203fd-linux-x86-64/contrib/swank-asdf.fasl" :VERBOSE NIL :PRINT NIL :IF-DOES-NOT-EXIST T :EXTERNAL-FORMAT :DEFAULT)
5: (REQUIRE :SWANK-ASDF #P"/home/alendvai/.slime/fasl/2015-11-23/sbcl-1.3.1.6.hu.dwim.4-ce203fd-linux-x86-64/contrib/swank-asdf.fasl")
6: (SWANK-REQUIRE (:SWANK-ASDF :SWANK-FANCY-INSPECTOR :SWANK-FUZZY :SWANK-INDENTATION :SWANK-PRESENTATIONS :SWANK-CLIPBOARD :SWANK-SPROF :SWANK-REPL :SWANK-C-P-C :SWANK-SBCL-EXTS) NIL)
7: (SB-INT:SIMPLE-EVAL-IN-LEXENV (FUNCALL (READ-FROM-STRING "swank:swank-require") (APPEND (QUOTE (:SWANK-ASDF :SWANK-FANCY-INSPECTOR :SWANK-FUZZY :SWANK-INDENTATION :SWANK-PRESENTATIONS :SWANK-CLIPBOARD :SWANK-SPROF :SWANK-REPL :SWANK-C-P-C)) (WHEN (FIND :SBCL *FEATURES*) (QUOTE (:SWANK-SBCL-EXTS))))) #<NULL-LEXENV>)
8: (SB-IMPL::SIMPLE-EVAL-PROGN-BODY ((REQUIRE :ASDF) (LOAD "/home/alendvai/common-lisp/slime/swank-loader.lisp" :VERBOSE T) (FUNCALL (READ-FROM-STRING "swank-loader:init")) (FUNCALL (READ-FROM-STRING "swank:swank-require") (APPEND (QUOTE (:SWANK-ASDF :SWANK-FANCY-INSPECTOR :SWANK-FUZZY :SWANK-INDENTATION :SWANK-PRESENTATIONS :SWANK-CLIPBOARD :SWANK-SPROF :SWANK-REPL :SWANK-C-P-C)) (WHEN (FIND :SBCL *FEATURES*) (QUOTE #)))) (FUNCALL (READ-FROM-STRING "swank:start-server") "/tmp/slime.13133")) #<NULL-LEXENV>)
9: (SB-INT:SIMPLE-EVAL-IN-LEXENV (PROGN (REQUIRE :ASDF) (LOAD "/home/alendvai/common-lisp/slime/swank-loader.lisp" :VERBOSE T) (FUNCALL (READ-FROM-STRING "swank-loader:init")) (FUNCALL (READ-FROM-STRING "swank:swank-require") (APPEND (QUOTE (:SWANK-ASDF :SWANK-FANCY-INSPECTOR :SWANK-FUZZY :SWANK-INDENTATION :SWANK-PRESENTATIONS :SWANK-CLIPBOARD :SWANK-SPROF :SWANK-REPL :SWANK-C-P-C)) (WHEN (FIND :SBCL *FEATURES*) (QUOTE #)))) (FUNCALL (READ-FROM-STRING "swank:start-server") "/tmp/slime.13133")) #<NULL-LEXENV>)
10: (EVAL (PROGN (REQUIRE :ASDF) (LOAD "/home/alendvai/common-lisp/slime/swank-loader.lisp" :VERBOSE T) (FUNCALL (READ-FROM-STRING "swank-loader:init")) (FUNCALL (READ-FROM-STRING "swank:swank-require") (APPEND (QUOTE (:SWANK-ASDF :SWANK-FANCY-INSPECTOR :SWANK-FUZZY :SWANK-INDENTATION :SWANK-PRESENTATIONS :SWANK-CLIPBOARD :SWANK-SPROF :SWANK-REPL :SWANK-C-P-C)) (WHEN (FIND :SBCL *FEATURES*) (QUOTE #)))) (FUNCALL (READ-FROM-STRING "swank:start-server") "/tmp/slime.13133")))
11: (SB-EXT:INTERACTIVE-EVAL (PROGN (REQUIRE :ASDF) (LOAD "/home/alendvai/common-lisp/slime/swank-loader.lisp" :VERBOSE T) (FUNCALL (READ-FROM-STRING "swank-loader:init")) (FUNCALL (READ-FROM-STRING "swank:swank-require") (APPEND (QUOTE (:SWANK-ASDF :SWANK-FANCY-INSPECTOR :SWANK-FUZZY :SWANK-INDENTATION :SWANK-PRESENTATIONS :SWANK-CLIPBOARD :SWANK-SPROF :SWANK-REPL :SWANK-C-P-C)) (WHEN (FIND :SBCL *FEATURES*) (QUOTE #)))) (FUNCALL (READ-FROM-STRING "swank:start-server") "/tmp/slime.13133")) :EVAL NIL)
12: (SB-IMPL::REPL-FUN NIL)
13: ((LAMBDA NIL :IN SB-IMPL::TOPLEVEL-REPL))
14: (SB-IMPL::%WITH-REBOUND-IO-SYNTAX #<CLOSURE (LAMBDA NIL :IN SB-IMPL::TOPLEVEL-REPL) {10070648FB}>)
15: (SB-IMPL::TOPLEVEL-REPL NIL)
16: (SB-IMPL::TOPLEVEL-INIT)
17: ((LAMBDA NIL :IN UIOP/IMAGE:RESTORE-IMAGE))
18: (UIOP/IMAGE:CALL-WITH-FATAL-CONDITION-HANDLER #<CLOSURE (LAMBDA NIL :IN UIOP/IMAGE:RESTORE-IMAGE) {10070614DB}>)
19: ((FLET #:WITHOUT-INTERRUPTS-BODY-82 :IN SB-EXT:SAVE-LISP-AND-DIE))
20: ((LABELS SB-IMPL::RESTART-LISP :IN SB-EXT:SAVE-LISP-AND-DIE))

0] 
```

there's some conditional upgrade magic in swank-asdf, maybe that's what breaks.

attila-lendvai:
simpler: @fare, you forgot to drop asdf2 from here, no?

```
    #+(or asdf2 asdf3 sbcl ecl) swank-asdf
```

luismbo:
@attila-lendvai I believe @fare's goal was not to drop ASDF 2 support altogether but only ASDF < 2.26.

fare:
Yes, I wanted to preserve compatibility with ASDF 2.26, because that's what Quicklisp still comes with, even though all recent implementations now have ASDF 3 (and all have 3.1, except CLISP). Now, if the SLIME maintainers feel comfortable with dropping support for ASDF 2.26, I'm all for it!

I didn't see your failure, but maybe that's because my ASDF 3.1.6.4 comes as an upgrade on top of the implementation's 3.1.5. I'll investigate — sorry, not much time for hacking right now.

fare:
OK, I could reproduce the error.

And now I see the incorrect assumption made by swank-asdf: that any of the compatibility functions listed in there will forever exist. Oops.

Fixed in ASDF 3.1.6.6: https://gitlab.common-lisp.net/asdf/asdf/commit/413c0a0d49d2d9237d5c22fe34864549d61d3357

fare:
Regarding the line in swank-loader:
```
    #+(or asdf2 asdf3 ecl sbcl) swank-asdf
```
This is mostly non-sensical. All implementations that provide asdf now provide asdf3 (except mocl, unmaintained, and unsupported by swank) but not by the time swank-loader is typically loaded, asdf is not loaded, so you can't tell which version is installed, or whether one is (in case someone uses an antique CL). SBCL and ECL have long provided ASDF, but so has CCL, and so have many implementations; why should only these two get swank-asdf?

I'd personally include swank-asdf on all supported implementations; actually, even obsolete implementations that don't provide asdf at all can have asdf installed on top of them: gcl (unsupported by swank), scl, corman, etc. And asdf3 can be installed on top of all ancient versions of CL (except ABCL, which has a recent ASDF anyway).

I would personally encourage SLIME maintainers to assume that ASDF3 is always present, chastise implementations that don't have it (and/or tell users to install ASDF3 on top), and simplify swank-loader by making it rely on UIOP and ASDF. I'd say, gc, scl and corman are mostly dead anyway (so is clisp, sadly), and if someone cares to re-release them, he may as well include ASDF3.1 with them.

fare:
This updated swank-asdf modifies the asdefs backward-compatibility macro so that in the future, it can deal more gracefully with backward-compatibility functions disappearing.

luismbo:
@fare right now the user has to ensure ASDF is loaded before loading SWANK if she wants to get the slime-asdf contrib loaded. (Unless on CCL and SBCL where we simply assume proper ASDF will be there.) A better solution would be to try and softly load ASDF beforehand. A worse solution would be to fail if ASDF is not recent enough, since slime-asdf is not a critical contrib. This patch is doing the latter for ASDF in [2.00; 2.26[.

Assuming ASDF 3 is available is not an option. Bundling it as a fallback, Quicklisp style, might be.

fare:
I can see many advantages and disadvantages to distributing asdf. There are many ways of doing it, too: a copy, git submodules, etc. If that can help simplify the code and drop backward compatibility, though, that would make me happy.

luismbo:
@fare pursuing that route will take longer, of course, since there are many more details to iron out. :-) What do you want to do with this concrete pull request?

fare:
I think this pull request is progress and should be adopted to make SLIME more future-proof.
I updated the commit message to explain what the patch does.

luismbo:
@fare OK. My objection (and my proposed fix) to this pull request still stands.

fare:
Wait, what is the objection, and what is the fix?

If you want to improve swank-loader, that's great, but an independent issue.

I can send a second patch to swank-loader, that makes it always try to load asdf, and only load swank-asdf if recent enough, if you want.

luismbo:
@fare Suppose I've got an old version ASDF2 that is earlier than ASDF 2.26. If I load SLIME with your patch applied, it'll signal an error and fail to load SWANK altogether. My proposed fix is to not attempt to load `swank-asdf.lisp` if ASDF's version is not >= 2.26.

fare:
@luismbo well suppose you have an old version of ASDF1 or 2 that is earlier than ASDF 2.014.6; with the current SLIME you lose just as much. How far do you want to go?

When I wrote the current backward compatibility support, it was in January 2013, and ASDF 2.014.6 from April 2011 was 21 months old. 2.26 from October 2012 is 38 months old, and ASDF 3.1.2 from May 2014 is 20 months old. Maybe we should stop supporting anything older than ASDF 3.1.2? I'd say Requiring 3.0.1 wouldn't be too much, and allowing 2.26 is quite generous. All maintained implementations provide 3.1.2 except CLISP that provides 2.33 (the last ASDF 3 pre-release before 3.0).

fare:
Also, if we're going to load ASDF anyway just to check whether we'll load swank-asdf, we may as well load ASDF always and rely on UIOP rather than reinvent half of it in swank-loader (OK, actually UIOP "stole" a lot of code from swank-loader; but improved on it subsequently).

luismbo:
@fare point taken about the current situation with ASDF < 2.014.6. swank-asdf is not loaded if ASDF 1 is present (modulo sbcl/ccl). Can we do something like

```lisp
(ignore-errors
 (funcall (find-symbol (string '#:version-satisfies) :asdf)
          (symbol-value (find-symbol (string '#:*asdf-version*) :asdf))
          "2.26"))
```

Why am I (and @capitaomorte) being particularly picky about this? Because we have to deal with old ACL 8.1 and ACL 9.0 installations on a regular basis at work. I will check tomorrow what version of ASDF ACL 9.0 ships with.

fare:
I'm all for supporting old implementations (we support Genera, furgossake), but that's very different from supporting an old ASDF. ASDF was designed to be self-upgradable precisely so that the way to deal with an old ASDF is to upgrade it.

luismbo:
I see your point. I'm just worried that people that don't care about ASDF, who are the most likely to have an old ASDF laying around, will be unable to load SWANK even though they (obviously) don't care about the slime-asdf contrib. Does that make sense?

capitaomorte:
@fare, please consider @luismbo's point. Many people where we work with still use SLIME+Allegro 8.1 and *not* quicklisp or anything else that automatically updates ASDF.  They don't care about slime-asdf or asdf. 

I personally don't care, and one of the reasons I designed SLY to use on-demand contribs was to avoid precisely these problems. BTW that code, that we wrote together, is still lying in a SLIME branch somewhere...

fare:
I can send a separate patch to swank-loader.

How hard do you want it to load asdf? And at which point in the loader.

At some point, yes, it makes sense to bundle a copy of asdf with swank (and probably separate uiop and asdf/defsystem, too). Do you want me to do that?

fare:
OK, so now all implementations (including the last laggard, LispWorks) have adopted ASDF 3. (Actually, all of them provide ASDF 3.1, though CLISP hasn't been released with it yet.)

Can we declare ASDF < 2.26 dead? Actually, if we can also declare 2.26 dead and say everyone has ASDF 3, that's even better.

In any case, please consider this patch again.

Also, `operation-forced` is deprecated and will be removed from ASDF at some point. Please don't use it in swank-loader. Patch included.

luismbo:
Cleaning old compatibility code is good, but I'd prefer if we would abstain from loading swank-asdf if ASDF is not recent enough rather than signalling an error that causes SWANK to not load altogether. (Perhaps the solution is to signal some sort of `missing-dependency` error and the swank loader can handle causing it to skip the module in question.)

fare:
Where would the code for that be?

In swank-asdf.lisp, that loads ASDF, and would only load a swank-asdf-support.lisp if ASDF is recent enough?

In slime-asdf, doing some magic?

fare:
Where would the code for that be?

In swank-asdf.lisp, that loads ASDF, and would only load a swank-asdf-support.lisp if ASDF is recent enough?

In slime-asdf, doing some magic?

If you can propose pseudo code or more precise suggestion, I'm ready to implement it.

However, note that this is a new feature. Can we not block this merge in the name of such a new feature?

luismbo:
I don't have strong feelings about a specific solution. My idea is to tweak `compile-contribs` to handle a `missing-dependency` condition that `swank-asdf.lisp` can signal if ASDF is not new enough. Does that make sense?

This is indeed a new feature, but being able to load SLIME in the absence of ASDF3 (or whichever version we decide to require) is important for me.

fare:
It's a new feature -- can it go in a follow-up CR?

Who would define this condition, and who would catch it and do what with it?

You seem to know what you want and I'm not sure what it is. Can you do it, or at least give me some hints sufficient for me to write the code that will satisfy you?

ASDF 3 has been out for more than 3 years and ASDF 3.1 for more than 2 years. ASDF 3.1 is now ubiquitous.

attila-lendvai:
regarding that `:reload t` in the `swank.asd` file: won't that mean that swank will be unconditionally reloaded each time a `:depends-on (:swank)` is encountered somewhere, and thus undoing any redefinitions made by the swank contribs? took me a while to find out what on earth is going on...

any idea how to avoid that?

fare:
Instead of kluges, what about fixing swank to (1) use asdf properly, (2) use defvar instead of defparameter, etc., so that it can be properly reloaded without breaking extensions.

attila-lendvai:
FWIW, i support the idea of a proper `swank.asd` file, with or without deleting `swank-loader.lisp`.

fare:
I once wrote a proper swank.asd, but it was never merged.

I can re-write it if there's a chance of merge.

Also, considering all currently active implementations support asdf 3 out of the box, can we "just" use asdf instead of swank-loader?

attila-lendvai:
so, swank-loader.lisp could be one, completely ASDF-free way to load swank, and swank.asd could be another, orthogonal way of loading it.

and then the only remaining issue i see is to make sure that the swank-asdf contrib doesn't get loaded if ASDF is too old.

am i missing something?

(Stelian will open a PR for a proper swank.asd)

fare:
Yes. I believe at one  point I had a PR (this one?) that did that. But it has long bitrotten since.

luismbo:
@attila-lendvai right, I don't want SLIME to refuse to load if you happen to have an old ASDF laying around. I'm also curious about what this PR actually adds in terms of end-user features, but we didn't even get that far in this long discussion. :-)

fare:
How old is old? ASDF 1: 2004. ASDF 2: 2010. ASDF 3: 2013. ASDF 3.1: 2014. ASDF 3.3: 2017. Then there's mk-defsystem from 1991, and the original Lisp Machine DEFSYSTEM for the late 1970s.

All modern maintained implementations provide ASDF 3.1. On the antique ones that don't, you can install a recent ASDF (see asdf/tools/install-asdf.lisp).

Just who are you serving by extending backward compatibility too much? If they like antique software, shouldn't they be using an antique SLIME in an antique Emacs?

Because the downside is complexity, unmaintanability, barrier to implementing new features or to being adopted by newbies, etc.

stassats:
Older than it can cope with. And it's serving me.
> Because the downside is complexity, unmaintanability, barrier to implementing new features or to being adopted by newbies, etc.

That sounds like a description of ASDF.

fare:
Is ASDF more complex than it needs be? How would you refactor it better for the same functionality? Or if you believe some functionality can be done away with, which test in the regression test suite would you rather see fail for what simplification gained?

Have you heard of any new feature that anyone wanted to see implemented that wasn't due to ASDF not being simple or maintainable enough?

What newbies are put off by ASDF, and what alternative is there or would there be that is more attractive?

fare:
stassats: what implementation are you using that has an antique ASDF? How antique is that ASDF? If you can load a recent SLIME on it, is there a reason you cannot also load a recent ASDF on it?

attila-lendvai:
it kinda became a fashion to complain about contemporary ASDF, but back in the day when we were busy working on production systems, ASDF gave me a lot of headache by its unpredictable misbehavior. we even had bugs that seemed to be influenced by filesystem-order...

@fare fixed all of those bugs that i used to see. did it increase complexity? that depends on where you draw the line... because if we also include the mental model of how to compile a large system using the old ASDF, then the recent work on ASDF have certainly decreased complexity.

luismbo:
To be clear, I don't hate ASDF, at all. I use a LISPM-inspired DEFSYSTEM at
work and I sometimes wish I could use ASDF instead.

It's annoying to have to upgrade ASDF just to load SLIME when I'm not going
to use it. (We still have to interact with production systems
contemporaneous with ASDF1 once in a while.) To make matters worse, I'd be
on the receiving end of co-worker complains. 😅

attila-lendvai:
i know @luismbo, things have changed since those times.

re old ASDF: but why isn't it an option for people sticking to old code to also not update slime?

but if they load swank using a completely ASDF-free swank-loader.lisp, then it would solve this problem of backwards compatibility, too.

so, i'll wait for Stelian to open the PR for his swank.asd and then maybe look into a PR that implements this.

luismbo:
We interact with lisp images built using ACL 8, 9 and 10. Having to switch SLIME versions would be pretty annoying. Old ACL versions shipped with old ASDFs, so old deployments of our software have old ASDF versions built-in, which we don't use, anyway.

stassats:
Slime must work without ASDF or if an older ASDF is present. But swank.asd can do whatever.

PR #283, opened 2015-12-03
Class graph

jimka2001:
slime contrib to creating CLOS class graph using graphviz


easye:
Looks cool.

jimka2001:
:-) glad you like it.  I'm trying to figure out how to add some interactivity to the image.  Does anyone have a suggestion about how to map mouse events to elisp functions in a buffer displaying a png file.  If I could do that it could be much more flexible.    I'd need the elisp function to receive some sort of arguments describing where in the image the click was done.   As an alternative, I could provide callbacks for specific rectangles, corresponding to the class rectangles.

I'm happy to have suggestions from anyone who has an idea.

jimka2001:
Also the current implementation is very hostile with the window/frame.  It currently takes over the frame to display the image.  Please give me some suggestions of how to make this friendlier, both from a user-interface perspective and also from a code implementation perspective.

PuercoPop:
@jimka2001 idk how to make images interactive but there cedet comes with is [cogre](http://cedet.sourceforge.net/cogre.shtml) which uses picture-mode to build class graphs.

jimka2001:
This commit includes a cl-yacc based parser for the dot file, which calculates the coordinates
of all the class rectangles in the image.

luismbo:
@jimka2001 this looks pretty cool, although it seems you may be close to the edge of what's possible/idiomatic with Emacs. Good luck. :-)

Issue #200, opened 2014-10-24
Support expansion of macrolet in slime-macroexpansion mode


fstamour:
Feature request.

It would be very nice to be able to use the function `slime-expand-1-inplace` (default binding `C-c RET`) on macrolet when in the slime-macroexpansion mode.

I am not sure if I am explaining what I want correctly, if not, just ask for clarifications. And if somebody tips me on how it could be done, I could even try to do it myself.

Issue #157, opened 2014-04-28
MACROEXPAND-ALL for ECL


phmarek:
Hi,

since e6ad5fa90e472 ECL has the code-walker available again. Please enable WALKER:MACROEXPAND-ALL (perhaps conditionalized on the package existence).

Thank you.

PR #284, opened 2015-12-04
Macrostep contrib

joddie:
This is a hack @luismbo and I have been working on, which I hope is close to being ready for inclusion.  It provides a slightly fancier way of doing inline macro-expansion by integrating with the `macrostep` library (http://github.com/joddie/macrostep), originally written for working with Emacs Lisp.  The main features are:

- minor-mode for expanding and collapsing inline expansions in the source buffer
- syntax highlighting of macro and compiler-macro forms within the expansion
- expansion of locally-defined macros in context, as requested in issue #200 

If desired, the context-sensitive expansion feature could be factored out of this patch so that it could also be used by the builtin `slime-macroexpand-1-inplace`.

Except where noted, the included ERT tests pass under Emacs 24.4 on Mac OS X with Lisp implementations CMUCL, SBCL, CLISP, CCL and ABCL.  I have a patch which provides fuller functionality under ECL, but it requires issue #157 to be fixed.

Two (hopefully minimal) changes were made to the core SWANK machinery to support this (primarily for context-sensitive expansion):

- `macroexpand-all` needs to take an optional second `environment` argument for the initial macro-expansion environment.  Working implementations have been added for SBCL, CMUCL, CCL and ABCL.  Other implementations currently, ignore the optional argument, so they should function as before.

- `macroexpand-all` under ABCL, which previously just called `macroexpand`, is redefined to use the implementation's `EXT:MACROEXPAND-ALL`, which has seemingly been available since 2009 (see, e.g. http://armedbear-devel.common-lisp.narkive.com/0TR5ZaAt/macroexpand-all ).

### DIFF
--- a/swank/backend.lisp
+++ b/swank/backend.lisp
@@ -699,7 +699,7 @@ available."
         (and (consp form) (length=2 form)
              (eq (first form) 'setf) (symbolp (second form))))))
 
-(definterface macroexpand-all (form)
+(definterface macroexpand-all (form &optional env)
    "Recursively expand all macros in FORM.
 Return the resulting form.")
 
--- a/swank/backend.lisp
+++ b/swank/backend.lisp
@@ -727,42 +727,45 @@ NIL."
                    (values new-form expanded)))))
     (frob form env)))
 
-(defmacro with-collected-macro-forms ((var) instrumented-form &body body)
+(defmacro with-collected-macro-forms
+    ((forms &optional result) instrumented-form &body body)
   "Collect macro forms by locally binding *MACROEXPAND-HOOK*.
 
-Evaluates INSTRUMENTED-FORM, with *MACROEXPAND-HOOK* bound to a
-function which collects all forms that undergo macro-expansion into a
-list stored in VAR.  Then evaluates BODY, with VAR still bound to the
-list of forms."
-  (assert (symbolp var))
-  (assert (not (null var)))
-  (let ((real-macroexpand-hook (gensym))
-        (instrumented-macroexpand-hook (gensym)))
-    `(let* ((,real-macroexpand-hook *macroexpand-hook*)
-            (,var '())
-            (,instrumented-macroexpand-hook
-             (lambda (macro-function form environment)
-               (let ((result (funcall ,real-macroexpand-hook
-                                      macro-function form environment)))
-                 (unless (eq result form)
-                   (setq ,var (cons form ,var)))
-                 result))))
-       (let ((*macroexpand-hook* ,instrumented-macroexpand-hook))
-         ,instrumented-form)
-       ,@body)))
+Evaluates INSTRUMENTED-FORM and collects any forms which undergo
+macro-expansion into a list.  Then evaluates BODY with FORMS bound to
+the list of forms, and RESULT (optionally) bound to the value of
+INSTRUMENTED-FORM."
+  (assert (and (symbolp forms) (not (null forms))))
+  (assert (symbolp result))
+  `(call-with-collected-macro-forms
+    (lambda (,forms ,(or result (gensym))) ,@body)
+    (lambda () ,instrumented-form)))
+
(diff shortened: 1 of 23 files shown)


joddie:
Hm, clearly there is work to do here. Will update as I investigate the test failures.

luismbo:
@joddie Is it that Emacs 23 doesn't have pcase? I believe we're ready to drop Emacs 23 support now that Debian stable ships with Emacs 24.

joddie:
Yes, that and a couple of other functions/macros which I used without being careful about when they were introduced (e.g. `define-error`, introduced in Emacs 24.4, IIRC).

I've actually now patched macrostep.el to restore compatibility with older Emacs, so I could push that if you'd like to support Emacs 23 a while longer.

joddie:
I merged from origin/master in order to resolve a merge conflict and get Travis to run again.  This makes the history messy, so it's probably preferable if this is rebased or squashed before any final merge to master.

joddie:
Pushed a rebased branch that should be compatible with Emacs 23.

luismbo:
Great! Can I go ahead and merge it? I'll work on adding a chapter to the manual about slime-macrostep. I've added you as a committer for future maintenance and stuff.

joddie:
I'd like to make one more tweak to push before merging this (a bugfix from the upstream macrostep.el repository).  Almost done.

swank/backend.lisp
```diff
@@ -727,6 +727,43 @@ NIL."
...
+  (let ((real-macroexpand-hook (gensym))
+        (instrumented-macroexpand-hook (gensym)))
+    `(let* ((,real-macroexpand-hook *macroexpand-hook*)
+            (,var '())
+            (,instrumented-macroexpand-hook
+             (lambda (macro-function form environment)
+               (let ((result (funcall ,real-macroexpand-hook
+                                      macro-function form environment)))
+                 (unless (eq result form)
+                   (setq ,var (cons form ,var)))
```
luismbo: Minor refactoring suggestion: `(push form ,var)`.

Slightly less minor refactoring suggestion, move non-syntactic bits to a function.

```lisp
(defun call-with-collected-macro-forms (body-fn instrumented-form-fn)
  (let* ((real-macroexpand-hook *macroexpand-hook*)
	 (forms '())
	 (instrumented-macroexpand-hook
	   (lambda (macro-function form environment)
	     (let ((result (funcall real-macroexpand-hook
				    macro-function form environment)))
	       (unless (eq result form)
		 (push form forms))
	       result))))
    (let ((*macroexpand-hook* instrumented-macroexpand-hook))
      (funcall instrumented-form-fn))
    (funcall body-fn forms)))

(defmacro with-collected-macro-forms ((var) instrumented-form &body body)
  "Collect macro forms by locally binding *MACROEXPAND-HOOK*.

Evaluates INSTRUMENTED-FORM, with *MACROEXPAND-HOOK* bound to a
function which collects all forms that undergo macro-expansion into a
list stored in VAR.  Then evaluates BODY, with VAR still bound to the
list of forms."
  (assert (symbolp var))
  (assert (not (null var)))
  `(call-with-collected-macro-forms (lambda (,var) ,@body)
				    (lambda () ,instrumented-form)))```

joddie:
Refactored more or less as per @luismbo's suggestion, and updated macrostep.el to latest version.  Probably best to squash once more before doing the final merge ...

luismbo:
Squashed the last two commits -- I hope that's what you meant and I hope I didn't screw the ChangeLog up -- and pushed this to master. Hurrah. Created issue #298 to keep track of the missing documentation.

Thanks again. :)

joddie:
Thanks Luís! Hope it is useful to you & others. I am very time-limited for the next month but I will do my best to work on the documentation issue after that.

joddie:
BTW, I don't see it on master yet, have you pushed the merge?

luismbo:
@joddie I fumbled the pushing into the wrong branch. Should be in master now. Thanks for the heads up.

PR #291, opened 2015-12-31
 Add support for UNIX domain sockets

drbo:
This adds support for UNIX domain sockets and makes them the only
option on platforms and Common Lisps that support them.

There are still bugs:

- The temporary files are not deleted when SLIME exits.  Note that
  these will be deleted in the event of a PID clash.
- TCP sockets are still completely insecure.

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -175,12 +175,14 @@ Return nil if the ChangeLog file cannot be found."
 This applies to buffers that present lines as rows of data, such as
 debugger backtraces and apropos listings."
   :type 'boolean
+  :risky t
   :group 'slime-ui)
 
 (defcustom slime-kill-without-query-p nil
   "If non-nil, kill SLIME processes without query when quitting Emacs.
 This applies to the *inferior-lisp* buffer and the network connections."
   :type 'boolean
+  :risky t
   :group 'slime-ui)
 
 ;;;;; slime-lisp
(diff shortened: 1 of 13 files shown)


drbo:
Temp files deletion fixed, but I can't understand why I broke the test suite.

drbo:
Test suite fixed.  `~/.slime-secret` support is broken though.

DemiMarie:
I just realized a problem: this relies on Emacs for temporary file handling.  So it won't work in the case of a remote Swank being started outside of Emacs.

luismbo:
@DemiMarie do we fallback to TCP in that case?

DemiMarie:
This patch doesn't have a fallback, for security reasons.  I can remove the
dependency when done with an exam.

For remote debugging, my plan is to use SSH to handle communication.

luismbo:
Fallback was a bad choice of words. I meant, can we pass a specific `:tcp` or `:unix` option to create-server? `:unix` would be the default, and SLIME can invoke it with `:unix` when applicable.

BTW, creating a local swank server is useful for situations other than remote debugging.

slyrus:
Any chance of getting this pulled soon?

luismbo:
This pull request removes the ability to open a SWANK server using TCP on Unix. I don't think that's reasonable. It needs be updated so that users can choose whether they TCP or UNIX_AF. I'm fine with SLIME defaulting to opening using UNIX_AF by default, when available.

slime.el
```diff
@@ -195,27 +197,32 @@ This applies to the *inferior-lisp* buffer and the network connections."
...
   :group 'slime-lisp)
 
 (defcustom slime-lisp-host "127.0.0.1"
   "The default hostname (or IP address) to connect to."
   :type 'string
+  :risky t
   :group 'slime-lisp)
 
-(defcustom slime-port 4005
+(defcustom slime-port "~/.slime/socket"
```
cwebber: This probably works; notably what emacs' server functionality does is creates a socket file at `/tmp/{username}{userid}/server`, which is another way to go

slime.el
```diff
@@ -1291,13 +1350,12 @@ The default condition handler for timer functions (see
     (cancel-timer slime-connect-retry-timer)
     (setq slime-connect-retry-timer nil)))
 
-(defun slime-read-swank-port ()
+(defun slime-read-swank-port (slime-swank-port-file)
   "Read the Swank server port number from the `slime-swank-port-file'."
```
cwebber: Seems like this docstring should be changed

jcguu95:
What's the current status of this PR?

DemiMarie:
> What's the current status of this PR?

I am no longer working on it as I quit using Common Lisp years ago.  If someone else wants to work on it, that would be great!

PR #293, opened 2016-01-04
some changes that i propose for inclusion into master

attila-lendvai:


attila-lendvai:
i'm wondering if that failure is genuine or just a transient... and if genuine, then how can i debug it, because the output is not too useful.

luismbo:
Most likely transient since it was ccl-only. Retried.

Issue #294, opened 2016-01-28
Can't work with kawa


s-kostyaev:
% java -cp /usr/share/java/kawa.jar:/home/feofan/.emacs.d/elpa/slime-20160113.630/contrib/swank-kawa.jar:/usr/lib/jvm/java-8-openjdk/lib/tools.jar -agentlib:jdwp=transport=dt_socket,server=y,suspend=n -Xss2M kawa.repl --r7rs -s
Listening for transport dt_socket at address: 35601
#|kawa:1|# (import (swank-kawa))
#|kawa:2|# (create-swank-server 4005)
Listening on port: 4005
connection: Socket[addr=/127.0.0.1,port=47278,localport=4005]
listener: Thread[swank-listener,5,main] 1308444215 chan@207abc3 #<environment main>
attaching: 6345 20
attaching2: com.sun.jdi.ProcessAttach (defaults: pid=, timeout=) {pid=pid=6345, timeout=timeout=20}
canBeModified: #t
canRedefineClasses: #t
canAddMethod: #f
canUnrestrictedlyRedefineClasses: #f
canGetBytecodes: #t
canGetConstantPool: #t
canGetSyntheticAttribute: #t
canGetSourceDebugExtension: #t
canPopFrames: #t
canForceEarlyReturn: #t
canGetMethodReturnValues: #t
canGetInstanceInfo: #t
closing socket: ServerSocket[addr=0.0.0.0/0.0.0.0,localport=4005]
Exception in thread "swank" java.lang.Error: mcase failed tmp
[...]

PR #295, opened 2016-01-28
fix swank-kawa #294

s-kostyaev:


luismbo:
Nice job. Do you have any idea why this changed? Perhaps it has to do with JVM or Kawa versions?

s-kostyaev:
I think it may be kawa version, because I was try slime & kawa with various jvm versions and have issue on both. I need to read kawa changelog for fully understanding.

luismbo:
Should you not reach a definitive conclusion, perhaps the following conservative approach would work well?

```scheme
(let* ((class :: <class-type> (1st (append (! classesByName vm "swank-kawa")
                                           (! classesByName vm "swank$Mnkawa"))))
       ...)
  ...)
```

s-kostyaev:
Backward compatibility is very good idea.

s-kostyaev:
But I can't do this:
error Argument #1 '[]' to 'append' has wrong type (java.util.Collections$UnmodifiableRandomAccessList)

luismbo:
Uh, oh, seems like your Java fears came true. Can you figure out how to append two collections?

s-kostyaev:
I can fix it without Java black magic :)

luismbo:
Hmm, try `vector-append` instead of `append`.

s-kostyaev:
With `vector-append` I have the same exception. Do you think that my current solution is bad?

luismbo:
No, it's fine. Appending would be slightly more elegant, that's all.

s-kostyaev:
For now I don't know how to append `Unmodifiable` with kawa :)

luismbo:
No problem. Applied your fix in 0525b90, thanks!

PR #300, opened 2016-02-12
improve interaction with kawa repl

s-kostyaev:
This PR add evaluation in the parent's environment (in this env work repl in `*inferior-lisp*` buffer). I think this may improve interaction with kawa repl in slime. For example, if you write new procedure definition in slime buffer and call `slime-eval-last-expression` you may switch to `*inferior-lisp*` buffer and interact with this procedure.


s-kostyaev:
I can use `(slime-setup '(slime-repl))` in `init.el` instead.

PR #304, opened 2016-03-09
Fix simple read

zellerin:
Make `simple-read` (and `*validate-input*`) really work.

I have it on for some time and no issue. There may be later interactions, still.


luismbo:
Pushed with a little tweak to the LOOP indentation, because OCD. Thanks!

Issue #302, opened 2016-03-08
slime-secret protection can be bypassed


zellerin:
The initial connection slime-secret mechanism reads input as a form. This means that even without knowledge of the secret one can execute any command (e.g., pass it `000019(:hello #.(cl:print 'Hi))` as first input to connection instead of secret), or bypass the protection completely (pass it `000017#.(swank::slime-secret)` instead of secret).

The most direct solution is to read, if secret is required, the known number of octets into sequence and just compare them without any interpretation.

I can make a patch if this approach is ok.

luismbo:
That makes sense. I wonder if setting `swank::*validate-input*` to true would be the proper fix for that?

zellerin:
Setting `swank/rpc::*validate-input*` to `T` makes my next repl statement to fail immediately and close connection. And slime would not start after that. Current form of `simple-read` apparently (as the first issue) fails to interpret `:` as a package separator, and keywords are present there. And maybe - just maybe - use of read macros would make that function cleaner.

But if implemented well, it could prevent this particular attack. What appears to immediately work and should fix the issue at hand is locally setting `swank/rpc::*validate-input*` around the call to `decode-message` in `authenticate-client`:
```lisp
(defun authenticate-client (stream)
  (let ((secret (slime-secret))
        (swank/rpc::*validate-input* t)) <-- added
    (when secret
      (set-stream-timeout stream 20)
      (let ((first-val (decode-message stream)))
        (unless (and (stringp first-val) (string= first-val secret))
          (error "Incoming connection doesn't know the password.")))
      (set-stream-timeout stream nil))))
```
On the other hand, any incremental manipulation of the string before it is compared to expected text increases chance of possible mischief.

mpsyco:
Maybe set [\*read-eval*](http://www.lispworks.com/documentation/HyperSpec/Body/v_rd_eva.htm#STread-evalST) to nil?

zellerin:
`*read-eval*`: this is an obvious possibility, but blacklisting individual offending constructs usually means, from my experience, that you forget something (e.g., do we care about local attacker doing DoS attacks on you? If so, is `#1#` a problem? etc). Whitelisting - even if form of current `validate-input` - says more precisely what we expect. (Other thing is what we do when the error is raised - currently, slime does not handle malformed input too well)

But for the password: we know what we exactly expect as first chunk. We know how many octets it should have. So the safest possibility is to just read so many octets and compare that.

luismbo:
[...]

PR #305, opened 2016-03-15
Improve passwords handling

zellerin:
This is a set of simple fixes to increase security of password handling.

They probably do not conform to the best coding practices (in particular, use internal symbols of other packages), but show some problems with current situation. Feedback welcome. See also #302.

Was tested only on sbcl/Linux.

The goal is for the server to survive things like nmap -sV or random connection to the port, not sofisticated attacker going for DoS.

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -781,7 +781,8 @@ first."
       (send-to-sentinel `(:stop-server :socket ,socket)))))
 
 (defun authenticate-client (stream)
-  (let ((secret (slime-secret)))
+  (let ((secret (slime-secret))
+        (swank/rpc::*validate-input* t))
     (when secret
       (set-stream-timeout stream 20)
       (let ((first-val (decode-message stream)))
--- a/swank.lisp
+++ b/swank.lisp
@@ -775,20 +775,29 @@ first."
                                                :buffering t)
                   (unless dont-close
                     (close-socket socket)))))
-    (authenticate-client client)
-    (serve-requests (make-connection socket client style))
-    (unless dont-close
-      (send-to-sentinel `(:stop-server :socket ,socket)))))
+    (when  (authenticate-client client)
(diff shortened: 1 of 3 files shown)


swank.lisp
```diff
@@ -775,19 +775,29 @@ first."
...
+            (progn
+              (set-stream-timeout stream 20)
+              (let ((first-val (decode-message stream)))
+                (unless (and (stringp first-val) (string= first-val secret))
+                  (error "Incoming connection doesn't know the password.")))
+              (set-stream-timeout stream nil)
+              t)
+          (error (e)
+            (close stream)
+            (warn "Swank password: ~a" e)
```
luismbo: Should this say instead `Swank authentication failed: `?

zellerin: Yes, this is probably better wording.

swank.lisp
```diff
@@ -775,19 +775,29 @@ first."
...
-          (error "Incoming connection doesn't know the password.")))
-      (set-stream-timeout stream nil))))
+  "NIL if secret password was expected but not provided correctly."
+  (let ((secret (slime-secret))
+        (swank/rpc::*validate-input* t))
+    (if secret
+        (handler-case
+            (progn
+              (set-stream-timeout stream 20)
+              (let ((first-val (decode-message stream)))
```
svetlyak40wt: Changelog saying: Read password message in more secure way.

Why this is more secure if you leave decode-message here? Shouldn't it be replaced as suggested in #302?

svetlyak40wt: Changelog saying: Read password message in more secure way.

Why this is more secure if you leave decode-message here? Shouldn't it be replaced as suggested in #302?

zellerin: The same function behaves (a bit) more securely with `swank/rpc::*validate-input*` bound to `t`.

`decode-message` -> `read-message` -> `read-form`

PR #308, opened 2016-03-22
Remove ChangeLog files

luismbo:
This pull request removes SLIME's ChangeLog files. The rationale for this is twofold:

1. ChangeLog files introduce merge conflicts all too often. While Gnulib's ChangeLog merge driver attempts to solve that problem, installing and setting it up is troublesome and doesn't work on GitHub anyway.

2. While it is certaintly possible to write great ChangeLog entries, Emacs's ChangeLog being a notable example, the ChangeLog format promotes describing *what* changed rather than *why*.


sionescu:
Good idea.

PuercoPop:
Maybe include a section on commit messages in Contributing.md. Without a changelog commit messages are more important.

attila-lendvai:
let's say bye to that one last unnecessary merge conflict that, ironically enough, shows up above at the patch that removes the ChangeLog file... :)

stassats:
Nobody reads them anyhow. Good riddance.

luismbo:
@PuercoPop makes sense! I'll include a link to http://chris.beams.io/posts/git-commit/. Would you recommend a different article perhaps? Feel free to suggest specific verbiage too.

luismbo:
@attila-lendvai heh, it's an actual bug rather than a merge conflict, but, yeah, I think one last ChangeLog merge conflict will show up in this merge request, eventually. :-)

luismbo:
Finally pushed this to master!

PR #309, opened 2016-03-30
Make SWANK/BACKEND:COMPILER-MACROEXPAND-1 handle FUNCALL forms.

DalekBaldwin:
Technically the CL spec says only forms that look like
`(funcall (function name) ...)` will be expanded, but at least SBCL will also
expand calls of the form `(funcall (quote name) ...)`


swank/backend.lisp
```diff
@@ -718,8 +718,13 @@ defined, invoke the expander function using *macroexpand-hook* and
...
   (let ((fun (and (consp form)
-                  (valid-function-name-p (car form))
-                  (compiler-macro-function (car form)))))
+                  (or (and (eq (car form) 'funcall)
+                           (or (eq (caadr form) 'quote)
+                               (eq (caadr form) 'function))
+                           (valid-function-name-p (cadadr form))
+                           (compiler-macro-function (cadadr form)))
+                      (and (valid-function-name-p (car form))
+                           (compiler-macro-function (car form)))))))
```
luismbo: Perhaps this test can be refactored by first extracting the function name from the form? Maybe use an auxiliary function and call it `form-function-name` or something?

luismbo:
Drop the trailing period in the commit title, maybe make it `Handle FUNCALL forms in SWANK/BACKEND:COMPILER-MACRO-EXPAND-1`?

Also, the comment you added in the pull request is useful, perhaps it deserves to be included in the commit message?

DalekBaldwin:
Github auto-generated the PR message from the commit message, so it's already there.

PR #311, opened 2016-04-10
* contrib/swank-mit-scheme.scm: stop using `.com` extension explicitl…

venantius:
…y for load

Changes the compilation result filename to not explicitly use `.com` extensions
and instead allows the compiler to do its own extension lookup when calling
`load`. MIT scheme compiled for OS X will compile `.so` files but not `.com`
files.


luismbo:
Thanks!

PR #312, opened 2016-04-12
* mit-scheme: use correct function for file load

venantius:
The previous commit (c7726e03) failed to change `pathname-new-type` to
`pathname-name` when it removed the `com` extension. The function change
is necessary since we're not actually changing the filetype.

Sorry for not doing this properly in the previous commit.


luismbo:
@vevantius no problem. If you feel like it, adding MIT Scheme to the test suite would be welcome.

luismbo:
Thanks!

PR #315, opened 2016-04-16
contrib/slime-c-p-c.el: completion bugfix

atomontage:
+ Fixes "mapcar: Symbol's function definition is void:
  append-char-syntax" error when trying to complete #\

  This bug was introduced by Joao Tavora in:
  3467ebc9ead3408c21cd1d2ae119ac05e2a7e2d9


luismbo:
Thanks!

PR #318, opened 2016-05-23
include name of history file in "Continue?" prompt

jimka2001:
If there is an error loading the "history" file, emacs/slime gives me a prompt saying "some error" 
while loading history. Continue?"  But it fails to tell me which file it is talking about.

This PR simply updates the prompt to let me know what file it is talking about so I can take a
look at the file fix the problem.

### DIFF
--- a/contrib/slime-repl.el
+++ b/contrib/slime-repl.el
@@ -1173,12 +1173,12 @@ truncated.  That part is untested, though!"
 (defun slime-repl-safe-save-merged-history ()
   (slime-repl-call-with-handler
    #'slime-repl-save-merged-history
-   "%S while saving the history. Continue? "))
+   (format "%%S while saving the history %S. Continue? " slime-repl-history-file)))
 
 (defun slime-repl-safe-load-history ()
   (slime-repl-call-with-handler
    #'slime-repl-load-history
-   "%S while loading the history. Continue? "))
+   (format "%%S while loading the history %S. Continue? "  slime-repl-history-file)))
 
 (defun slime-repl-call-with-handler (fun query)
   "Call FUN in the context of an error handler.


jimka2001:
I've created an issue documenting this.  https://github.com/slime/slime/issues/319

ioannad:
The OP says in #319 that he does not understand why this error message occurs, and I have never gotten it myself. However, the request seems reasonable to me, and the fix is straightforward. 

Any opinions on this?

luismbo:
I have seen it happen to other users, but I don't yet know why it happens. Maybe it's related to having multiple Emacs instances? I'm not sure.

The only thing fishy about this change is that it's injecting an arbitrary string into a format string.

ioannad:
Would it be considered safer to first make sure that the file to which `slime-repl-history-file` points to, exists? (So it's not an arbitrary string?)

For example something like
```
(if (file-exists-p slime-repl-history-file)
    (format "%%S while saving the history %S. Continue? " slime-repl-history-file)
    "%S while saving the history, and the variable SLIME-REPL-HISTORY-FILE does not point to an existing file. Continue? ")
```

luismbo:
The case I was worrying about was if slime-repl-history-file has a funky name like `%special-file%` and the `%s` is interpreted as a format directive.

Maybe `slime-repl-call-with-handler` could not treat its argument as a format string?

```
(defun slime-repl-call-with-handler (fun query)
  "Call FUN in the context of an error handler.
The handler will use qeuery to ask the use if the error should be ingored."
  (condition-case err
      (funcall fun)
    (error
     (if (y-or-n-p (format "%s %s" (error-message-string err) query))
         nil
       (signal (car err) (cdr err))))))```

ioannad:
Oh, I see now, of course! :) 
It seems with your change, `slime-repl-call-with-handler` would be "cleaner", and arbitrary format strings cannot be injected to `format` anymore. 

Since @jimka2001 has not (yet?) responded (to perhaps include your proposed changes or give us permission to do so), I guess this could be merged when your change to `slime-repl-call-with-handler` gets merged. Would you make these changes? I would suggest fixing the typos in the documentation string as well, to 
  "Call FUN in the context of an error handler.
The handler will use the QUERY to ask the user if the error should be ignored."

jimka2001:
Hi everyone, about the unsantitized injection into the format string,  why not just sanitize the string `slime-repl-history-file` before passing it to format?   Do you really think changing the semantics of `slime-repl-call-with-handler` is a good idea?  To me it seems like overkill.

luismbo:
@jimka2001 `slime-repl-call-with-handler` is just a helper function for those two call sites.

haicnguyen:
Hi all,

The function `slime-repl-read-history' in file:contrib/slime-repl.el gave me a warning 

`"End of file during parsing" while loading the history. Continue? (y or n) y`

when I was `slime-connect'-ing and my history file ("~/.slime-history.eld") is totally empty. I found this warning was caused by the line "(read (current-buffer))" in the function `slime-repl-read-history' when the buffer it read from was empty.

(my SLIME version v2.24,  my GNU Emacs version 28.0.50 (development branch)).

PR #323, opened 2016-05-29
Allow slime to also encourage "sister" hackers

erjoalgo:


luismbo:
I'm definitely sympathetic towards the idea gender neutrality but we're paraphrasing a Beatles lyric here. Does it really make sense to tweak it this way?

luismbo:
Maybe

```lisp
(format "Take this REPL, %s, and may it serve you well." (slime-user-first-name))
```

would be slightly better (on average). :-)

(Which makes me think that one my subconscious objections was that calling SLIME users "sister" would be wrong most of the time. And that brings your point of being friendly to female hackers home!)

erjoalgo:
Even though it is a minor detail, having "brother" only made slime seem unfriendly and appear to cater only to male users, in my view, this is what I wanted to point out, but this fix is just an idea. Another option might be to add another witty line catering to eg females specifically, but 1. I couldn't think of any, and 2. it would still mean awkwardly trying to guess the users' gender, which might be better to avoid altogether.

So I would be in favor of your suggestion, `(format "Take this REPL, %s, and may it serve you well." (slime-user-first-name))`

. Maybe you can leave this open for a some time see what others think?

erjoalgo:
At the same time, this line has been a part of slime for a long time, and some users might "miss it", so I'm not sure even I am in favor of changing it.

taptap:
>and some users might "miss it"
Yes! I would not like to be addressed as `(slime-user-first-name)`, I would miss "brother" greeting.

>Another option might be to add another witty line catering to eg females specifically
This is good option, but it's not excluding the possibility to greet our sisters in REPL as well.

As for quoting Beatles... It's not a direct quote, anyway. 
And I personally always thought of it as a combination of an idea of global brotherhood of Lispers and fantasy-ish wish of luck with your endeavor.

luismbo:
@taptap I agree that removing "brother" would yank a fair amount of the phrase's essence. How would you feel if you were addressed with "Take this REPL, sister, and may it serve you well."?

luismbo:
An alternative to `(slime-user-first-name)` could be "friend".

taptap:
>An alternative to (slime-user-first-name) could be "friend".

The most neutral variant, but it also misses something, compared to sister/brother.

>How would you feel if you were addressed with "Take this REPL, sister, and may it serve you well."?

I would be slightly amused up to first ten or twenty times, I suppose. But then I wouldn't feel anything special compared to usual greetings. I can't tell for sure what exactly I would feel, of course.

luismbo:
Closing this PR since we have way too many open ATM. We can revisit this in the future if need be.

chee:
They had [this convo](https://github.com/clojure-emacs/cider/pull/415) over on clojure-emacs/cider too and [ended up with](https://github.com/clojure-emacs/cider/blob/1ed5163433c991c00ea83fdd4447e8daf4aeccbe/cider-util.el#L661) "fellow hacker". (though there were a couple of votes for comrade)

> How would you feel if you were addressed with "Take this REPL, sister, and may it serve you well."?
Better than i did about brother. I've overridden `slime-words-of-encouragement` in my `init.el` now though

stassats:
I'll just remove that line, easier than coming up with some convoluted phrases.

chee:
> I'll just remove that line, easier than coming up with some convoluted phrases.

Oh hey, thank you!

stassats:
But now, who even remembers Lemonodor?

lukego:
Hey neat, lemonodor is still online!

Here's the post from 2003 that introduced SLIME to a wider audience: http://lemonodor.com/archives/2003/09/slime.html.

K1D77A:
This is pathetic.

PR #324, opened 2016-05-30
Add mechanism for tagging Lisp connections.

wsgac:
I did some work to allow tagging of Lisp connections with custom tags. This comes in handy when there are multiple Lisps running (especially, multiple instances of the same implementation) and you want to be clear on which one is which. My changes accomplish three main goals:
- Allow tagging Lisp connections from the *slime-connections* buffer
- Allow setting a tag on Slime startup
- Display tags in Lisp file and REPL buffer modelines


luismbo:
Hi @wsgac. This is interesting. Are you dropping it because of lack of feedback or something else?

wsgac:
Hi @luismbo . No, actually today I looked at your Contributing document and in it you ask for topic branches. Besides, some tests were failing on Travis, so I decided I'll redo this in a less sloppy way. I created the previous pull request late at night and after that drifted away to other stuff.

luismbo:
OK, cool, sound good!

PR #328, opened 2016-06-29
Make fuzzy completion up/down navigation wrap around the edges

wsgac:
When using fuzzy completion I am frequently facing a fairly long completion list. I thought that not being able to wrap around the edges (i.e. scroll up from the beginning and arrive at the end and vice versa) was a bit counter-intuitive. So I modified the `slime-fuzzy-next` and `slime-fuzzy-prev` functions.

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,7 +26,7 @@ Quick setup instructions
      ```el
      ;; Set your lisp system and, optionally, some contribs
      (setq inferior-lisp-program "/opt/sbcl/bin/sbcl")
-     (slime-setup '(slime-fancy))
+     (setq slime-contribs '(slime-fancy))
      ```
 
   3. Use `M-x slime` to fire up and connect to an inferior Lisp. SLIME will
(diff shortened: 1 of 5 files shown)


attila-lendvai:
if you switch focus to the completion buffer, then you can jump to the beginning/end with the usual shortcuts. unfortunately the highlight is not updated, which should/could be fixed.

generally i'm not happy about such wrapping around, because it's often confusing for me, but this is just one opinion. i also rarely even look at anything one page down in the fuzzy list. if what i'm looking for is not at the top, then i usually type more and update the fuzzy list.

wsgac:
Well, yes, you can jump to the completion buffer, but to me the whole point of it intercepting navigation keys from another buffer is to make completion less of a cognitive distraction. I tested this change among some people at work and they liked it. Perhaps this behavior could be switched on/off with a configuration variable.

attila-lendvai:
don't add a config, that's pointless extra complexity.

if people generally like this change then just commit it as-is.

wsgac:
Yup, at least that's the impression I get by talking to other Lispers.

wsgac:
I believe there's nothing more to do here. Can you please merge this change?

PR #331, opened 2016-08-14
Fix byte-compiler warnings

### LINKED ISSUE #316: Compiler warnings and slime-enclosing-context
Hi, I'm getting some compiler warnings when byte-compiling SLIME:

```
In slime-parse-extended-operator-name:
slime-enclosing-context.el:45:36:Warning: reference to free variable
    `slime-extended-operator-name-parser-alist'
slime-enclosing-context.el:60:8:Warning: function
    `slime-parse-extended-operator-name' defined multiple times in this file
slime-enclosing-context.el:72:36:Warning: reference to free variable
    `slime-extended-operator-name-parser-alist'


In slime-lookup-reference-at-point:
slime-references.el:97:36:Warning: `common-lisp-glossary-fun' is an obsolete
    variable; use `common-lisp-hyperspec-glossary-function' instead.
```

They're easy to fix, but the one involving `slime-extended-operator-name-parser-alist'` is interesting.

Grepping the SLIME files gives no definition for this variable -- is it created in a different way? If it's never created, then it makes me wonder why the function `slime-parse-extended-operator-name` still exists if all it's doing is returning a few of its inputs.

This file was restored in #231

fice-t:
Fixes #316

### DIFF
--- a/contrib/slime-enclosing-context.el
+++ b/contrib/slime-enclosing-context.el
@@ -31,33 +31,6 @@ If SKIP-BLANKS-P is true, leading whitespaces &c are skipped.
       (member (char-syntax (aref string 0))
  '(?w ?_ ?\' ?\\))))
 
-(defun slime-parse-extended-operator-name (user-point forms indices points)
-  "Assume that point is directly at the operator that should be parsed.
-USER-POINT is the value of `point' where the user was looking at.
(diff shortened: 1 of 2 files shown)


fice-t:
@luismbo Does this look alright?

luismbo:
Yes, thanks! (Sorry for the delay.)

PR #335, opened 2016-08-31
Use the eldoc-highlight-function-argument face for consistency

dkim:
ElDoc mode has been using its own `eldoc-highlight-function-argument` face for the argument at point in a function's argument list.


luismbo:
Thanks.

PR #336, opened 2016-09-06
Enable goto source location to find definitions in jars

alanruttenberg:
Some source locations in ABCL are recorded as being in jar files. This patch modifies swank-abcl to recognize this situation and use the form (:zip file entry) for the file portion of the location spec.


alanruttenberg:
buggy.

PR #338, opened 2016-09-07
Enable goto source location to find definitions in jars

alanruttenberg:
Some source locations in ABCL are recorded as being in jar files. This patch modifies swank-abcl to recognize this situation and use the form (:zip file entry) for the file portion of the location spec.

### DIFF
--- a/swank/abcl.lisp
+++ b/swank/abcl.lisp
@@ -547,7 +547,7 @@
       (if (ext:pathname-jar-p path)
           `(:location
             ;; strip off "jar:file:" = 9 characters
-            (:zip ,@(split-string path "!/"))
+            (:zip ,@(split-string (subseq path 9) "!/"))
             ;; pos never seems right. Use function name.
             (:function-name ,(string symbol))
             (:align t))
--- a/swank/abcl.lisp
+++ b/swank/abcl.lisp
@@ -498,7 +498,7 @@
                        (not (load fn)))))))))
 
 (defimplementation swank-compile-string (string &key buffer position filename
-                                         policy)
+                                                policy)
   (declare (ignore filename policy))
   (let ((jvm::*resignal-compiler-warnings* t)
         (*abcl-signaled-conditions* nil))
@@ -506,8 +506,10 @@
       (let ((*buffer-name* buffer)
             (*buffer-start-position* position)
             (*buffer-string* string))
-        (funcall (compile nil (read-from-string
-                               (format nil "(~S () ~A)" 'lambda string))))
+        (let ((sys::*source* (make-pathname :device "emacs-buffer" :name *buffer-name*))
+              (sys::*source-position* *buffer-start-position*))
+          (funcall (compile nil (read-from-string
+                                 (format nil "(~S () ~A)" 'lambda string)))))
         t))))
 
 #|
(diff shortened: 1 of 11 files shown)


luismbo:
@easye can you review this?

swank/abcl.lisp
```diff
@@ -541,12 +541,19 @@
...
   (when (pathnamep (ext:source-pathname symbol))
     (let ((pos (ext:source-file-position symbol)))
-      `(:location
-        (:file ,(namestring (ext:source-pathname symbol)))
-        ,(if pos
-             (list :position (1+ pos))
-             (list :function-name (string symbol)))
-        (:align t)))))
+      (if (ext:pathname-jar-p (ext:source-pathname symbol))
+          (destructuring-bind (jar entry) (abcl-asdf::split-string (subseq (namestring (ext:source-pathname symbol)) 9) (char "!" 0))
```
luismbo: There's a `split-string` function in this file. Why are you grabbing it from abcl-asdf? Also, what does the `(subseq ... 9)` do? (i.e. what's it discarding?) Perhaps a comment or a clearer way of doing this (perhaps some pathname manipulation?) would be better?

swank/abcl.lisp
```diff
@@ -541,12 +541,19 @@
...
+                        ,(if nil ;pos -- positions seem to be wrong. Just use function name
+                            (list :position (1+ pos))
+                            (list :function-name (string symbol)))
+                        (:align t)))
+          `(:location
+            (:file ,(namestring (ext:source-pathname symbol)))
+            ,(if nil;pos
+                 (list :position (1+ pos))
+                 (list :function-name (string symbol)))
+            (:align t))))))
```
luismbo: There's a lot of duplication between these two location blocks. Seems like it should be easily refactored.

swank/match.lisp
```diff
@@ -158,17 +158,17 @@
         ((consp pattern)
          (append
           `((consp ,key))
-          (compile-select-tests (!cs-car key) (car
```
luismbo: Who's setting up a `!` reader macro?

luismbo:
This whole "changes requested" in bold red seems a bit harsh. :-) Anyway, let me know what you think of my comments. Cheers.

swank/match.lisp
alanruttenberg: I am. But this file wasn't part of my initial pull request. I don't understand how it got added...
This should be local to my fork. Not that I wouldn't mind have the change incorporated...

swank/abcl.lisp
alanruttenberg: split-string -> brain fart.
Subseq: Will add comment.
```
(describe #P"jar:file:/Users/alanr/repos/lsw/lib/abcl-contrib.jar!/jss/invoke.lisp")
#P"jar:file:/Users/alanr/repos/lsw/lib/abcl-contrib.jar!/jss/invoke.lisp" is an object of type JAR-PATHNAME:
  HOST         NIL
  DEVICE       (#P"/Users/alanr/repos/lsw/lib/abcl-contrib.jar")
  DIRECTORY    (:ABSOLUTE "jss")
  NAME         "invoke"
  TYPE         "lisp"
  VERSION      NIL
```
Using the pathname manipulations wouldn't be clearer since this is a nonstandard extension. I don't feel bad having this in the abcl-specific code.

swank/abcl.lisp
alanruttenberg: There are other places where code can be cleaned up. I haven't taken that on.

easye:
Just getting to this late.  You still want a review? I think Alan's code is better than the current situation.

swank/match.lisp
luismbo: I'm not sure about it. Seems like you shouldn't be messing with the standard readtable like that, but it doesn't hurt to get rid of those `!` characters...

swank/abcl.lisp
luismbo: Oh, sure, but this duplication was introduced by this pull request and seems easy to avoid. :-)

luismbo:
Thanks @easye. I think it's just needs some minor refactoring at this point and we're good to go!

alanruttenberg:
just added commit that tries to address comments, but also adds location tracking for functions defined in emacs buffers

alanruttenberg:
poke

easye:
@luismbo This is good to go from my side.  Shall I accept the merge request?

luismbo:
Cleaned up the PR into 3 commits and pushed to master. (Didn't test it, sorry. I hope I didn't break anything.) If in future PRs you could do that cleaning up, that'd be great, of course. Thanks!

alanruttenberg:
Thanks. I'm just getting used to the pull request process. Looked at your commits and will try to emulate this separation in the future.
Alan

PR #346, opened 2016-09-25
Add SLIME-INIT-COMMAND-HOOKS

phoe:
I want my Lisp image to `(setf *print-right-margin* 110)` whenever it's called from SLIME.

I thought of placing this code in my Lisp's RC file, but that would also call this in non-SLIME sessions, which I don't want to happen.

I found that there is no possibility to customize commands that are run whenever SLIME launches inferior-lisp.

This commit is meant to patch this possibility and allow people to customize the commands Lisp image is launched with when and only when Lisp is launched from SLIME.

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -1197,12 +1197,16 @@ See `slime-start'."
   (with-current-buffer (process-buffer process)
     slime-inferior-lisp-args))
 
+;; List of all hooks that are to be executed by Common Lisp image when it is
+;; called by SLIME.
+(defvar slime-init-command-hooks '())
+
 ;; XXX load-server & start-server used to be separated. maybe that was  better.
 (defun slime-init-command (port-filename _coding-system)
   "Return a string to initialize Lisp."
   (let ((loader (if (file-name-absolute-p slime-backend)
                     slime-backend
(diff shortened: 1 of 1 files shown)


phoe:
I discussed the issue over at `#lisp-pl` on Freenode and have a question. I'm used to hooklists in elisp being just bare variables, but a suggestion was to wrap the `slime-init-command-hooks` variable in a getter/setter of sorts to create an abstraction layer.

What do you think about it?

luismbo:
How about using `~/.swank.lisp`? https://common-lisp.net/project/slime/doc/html/Lisp_002dside.html

phoe:
Sounds good enough. Thanks!

luismbo:
OK, cool. I wonder if the slime-repl should manage `*print-right-margin*` based on the repl buffer's width? (That might be surprising for users, I guess. Just a random idea. ;-))

phoe:
Actually - yes, it should manage it. When a minor mode is toggled, there should be an Emacs hook for changing a buffer's width that checks if the current buffer is a Slime REPL - and if yes, it setfs the current width to the `*print-right-margin* of Lisp beneath.

PR #348, opened 2016-09-28
Spoof pathname variables when compiling from string.

rpgoldman:
This isn't a real pull request, but a possible direction for compiling from string in general.  I will file an issue related to this.

Add new CONTEXTUALIZED-CODE function that spoofs bindings for
*compile-file-pathname*, *compile-file-truename*, *load-pathname*, and
*load-truename*.

This makes compiling a single form behave more like when it was compiled
in context.

### DIFF
--- a/swank/allegro.lisp
+++ b/swank/allegro.lisp
@@ -522,12 +522,32 @@ to do this, this factors in the length of the inserted header itself."
       (write-string padding-string stream)
       (write-char #\newline stream))))
 
+(defun contextualized-code (string file)
+  `(let ((*compile-file-pathname* (or (eval-when (:compile-toplevel)
+                                        ,(parse-namestring file))
+                                      nil))
+         (*compile-file-truename* (or (eval-when (:compile-toplevel)
+                                        ,(truename (parse-namestring file)))
+                                      nil))
+         (*load-pathname* (or (eval-when (:load-toplevel)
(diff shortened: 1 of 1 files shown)


luismbo:
Makes sense! It'd be nice if we could standardise this sort of thing across all backends.

rpgoldman:
If this makes sense, I could pull it out of Allegro and put it into a package where all the backends could get access to it. Just wanted to make sure it wasn't objectionable.

rpgoldman:
I put it in place in all the lisps I could figure out, see my [allegro-compile-string branch](/rpgoldman/slime/tree/allegro-compile-string).

For some implementations there were methods of compiling strings that did not use a temporary file, and these methods I did not attempt to change.

This has not been extensively tested, but I make it available for comment and testing.

alanruttenberg:
Did this independently for abcl
https://github.com/alanruttenberg/slime/commit/30566b174be3186fde1ac224d4fffd3def0f4d9e
I'll submit a pull request when the abcl distribution catches up

rpgoldman:
I'm confused by the status of this pull request.  Yes, a travis run for this pull request failed -- specifically 791.1 - 791.6 failed.  But the last two Travis builds passed.  Yet this still shows as a broken pull request.  Will someone please clue me in about this?  Is there something I should do to repair this?  I'd like to see it merged and go away...

PR #350, opened 2016-10-04
Remove call to ASDF::OPERATION-FORCED.

rpgoldman:
Please try to merge this directly.

The function `OPERATION-FORCED` is being removed from ASDF, and at any rate was never an API function.

Previously, this function always returned `NIL`, but that behavior breaks reloading swank.  Rather than "fixing" it by changing the function to always return `T`, we are going to simply remove it.

### DIFF
--- a/swank.asd
+++ b/swank.asd
@@ -31,8 +31,8 @@
 (defmethod asdf:perform ((o asdf:load-op) (f swank-loader-file))
   (load (asdf::component-pathname f))
   (funcall (read-from-string "swank-loader::init")
-           :reload (asdf::operation-forced o)
-           :delete (asdf::operation-forced o)))
+           :reload nil
+           :delete nil))
 
 (asdf:defsystem :swank
   :default-component-class swank-loader-file


rpgoldman:
Blast. I misunderstood Fare's explanation I see, and got this precisely backwards. I will revise the pull request and resubmit.

rpgoldman:
OK, there it is.  Replaced calls to `OPERATION-FORCED` with `NIL`.

luismbo:
I'm confused about the commit message. Why does passing T causes SLIME to fail to reload?

rpgoldman:
Fare writes:

> As of 2016-10-01, SLIME doesn't actually want to be reloaded, and dies badly when swank-loader:init is called with :delete t. So have operation-forced return constantly nil instead of constantly t.

This is his comment for a patch to `OPERATION-FORCED`, but that patch isn't going to be applied: SLIME's swank.asd is the only use of this function, and we are going to remove it in the next ASDF release.

luismbo:
Oh, if you meant that `:delete t` will cause the current swank/slime connection to die, then yeah it'll die badly I guess. :-)

luismbo:
Tweaked the commit message a little bit. Thanks!

fare:
Personally, I'd have used `:delete nil :reload t` which I believe is more correct and more in line with what ASDF usually does. I had a patch to make `swank.asd` into an actual asd file, but it needs some rebasing and loving.

luismbo:
Makes sense. For some reason, I had mistakenly interpreted that `asdf:oos`'s `:force t` option was being removed. IIUC now, if SWANK is already loaded (via ASDF), unless we pass `:force t` to `asdf:load-system`, our `asdf:perform` method won't be invoked at all. That is why it didn't make sense to check for `operation-forced`.

a6c0110 implements your suggestion, thanks.

PR #351, opened 2016-10-14
Properly handle number specializers in eldoc help

wsgac:
I found that when I have a method with an `eql` specializer that happens to be a number, I (and some of my colleagues) notice an error similar to: `error in process filter: Wrong number of arguments: nil, 111`. One simple example triggering the error would be this:
```
(defmethod test ((a (eql 2)))
                      a)
```
The change in this push request solves the problem for number specializers.

### DIFF
--- a/contrib/swank-arglists.lisp
+++ b/contrib/swank-arglists.lisp
@@ -412,6 +412,7 @@ Otherwise NIL is returned."
   "ARG can be a symbol or a destructuring pattern."
   (etypecase arg
     (symbol        arg)
+    (number        arg)
     (arglist-dummy arg)
     (list          (decode-arglist arg))))
 


luismbo:
Interesting, this will fail with all sorts of objects, not just numbers. The issue seems to be a little bit upstream, in `SWANK::METHODS-KEYWORDS` which is delving too deep into the lambda list using `ARGLIST-KEYWORDS` which is unaware of defmethod lambda-list syntax.

Do you feel like improving `arglist-keywords` to not do that? It might be a matter of having `decode-arglist` not recurse into arguments. Not sure.

wsgac:
Sure. I can have a look at this over the weekend. I'll get back to you when I need to discuss the details.

PR #353, opened 2016-10-15
Fix 'set-default-directory' function

ivanp7:
Change 'namestring' to 'pathname' in 'set-default-directory' function because 'ext:chdir' accepts a filepath object, not a string


ivanp7:
ext:chdir was changed to accept strings

luismbo:
Thanks for the follow-up.

PR #354, opened 2016-10-17
Fuzzy completion sets itself as default completion when loaded.

attila-lendvai:
please consider merging this.

the reasoning: newcomers should be exposed to fuzzy completion, and people who don't like fuzzy can always customize it away at their liking.

### DIFF
--- a/contrib/slime-fuzzy.el
+++ b/contrib/slime-fuzzy.el
@@ -3,6 +3,8 @@
 (require 'slime-c-p-c)
 (require 'cl-lib)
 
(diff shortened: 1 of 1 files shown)


attila-lendvai:
any word on this?

who even feels having the authority to accept or reject this nowadays?

luismbo:
Yes, I agree. I have some completion improvements in the pipeline, and I'll enable fuzzy completion by default. Closing (at least for now) since our changes will likely collide. (Sorry.)

PR #364, opened 2016-12-15
fancy inspector improvements

alanruttenberg:
1) Adds some extra information for the symbol inspector - symbols with the same name in other packages
2) Labels, like "Function:" were plain text. Make them '(:label "Function") instead thereby telling the inspector to use the label face.
3) Add a :strong-view for elements in another style. I uses this in a java class inspector in abcl to highlight the function name in the longer string describing it. e.g.

"public abstract int java.io.InputStream.read() throws java.io.IOException" becomes
```elisp
`((:view ,method "public abstract int java.io.InputStream.") 
 (:strong-view ,method "read") 
 (:view ,method,  "() throws java.io.IOException"))
```

### DIFF
--- a/contrib/swank-fancy-inspector.lisp
+++ b/contrib/swank-fancy-inspector.lisp
@@ -9,6 +9,15 @@
 (eval-when (:compile-toplevel :load-toplevel :execute)
   (swank-require :swank-util))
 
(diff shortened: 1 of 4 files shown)


easye:
Presumably superceded by https://github.com/slime/slime/pull/376

PR #366, opened 2016-12-24
make defimplementation create named function

alanruttenberg:
As currently implement, defimplementation puts an flet-defined function on the plist of the name of the function. The problem with this is that, at least in ABCL, that flet function doesn't get compiled. In addition I'm not sure whether source location is generally kept - it isn't in ABCL. 
 
Instead, create a function using defun, with the name of the defined function the symbol interned in the implementation package, and put *that* function on the plist.


alanruttenberg:
Hmm. Didn't realize that the backend symbols are imported into the implementation package, and so my defun blows away the original. Will contemplate, and either amend with another commit or close.

alanruttenberg:
Never mind. Chose a way that is more specific to ABCL.

PR #368, opened 2017-01-21
Correct the customization type of lisp-loop-body-forms-indentation

dkim:
The `lisp-loop-body-forms-indentation` customization variable is of type `integer` but was wrongly specified as `boolean`.


luismbo:
Merged, thanks!

PR #370, opened 2017-03-03
Remove useless ,@()

phoe:


luismbo:
Can you confirm that the plist indents properly even without the slime-indentation contrib? I think that's what the `,@()` is for.

PuercoPop:
@luismbo if it is still needed ideally there should be a comment noting its purpose, the commit on which was introduced only says 'make it so'. Was this a common idiom?

PR #376, opened 2017-03-27
Alan Ruttenberg 201703a

easye:
This merge consolidates Alan Ruttenberg's work on non-abcl specific portions of SLIME.
@easye vouches that these changes work well with ABCL, (and other popular open CL implementations SBCL, CCL, ECL) when rebased against https://github.com/slime/slime/tree/master

These patches include:

FIXES
------

1. Fixes for searching for functions and methods

ENHANCEMENTS
------------------

1. Load swank matcher earlier so that it may be used by implementation-specific code (has architectural issues for SLIME loader vs. swank.lisp availablity to reuse between implementation specific code)

2. Refactor utilties used by implementation-specific inspector code into swank/backend.lisp

3. Add an inspector :strong-view for a view with a bolder face

4. Strings may edited in an emacs buffer from the inspector 

5. Return information for definitions even if the symbol isn't found in the current package if the ```*find-definitions-all-packages*``` special  is set to non-nil.  By default it is set to nil, but currently the SLIME inspector still offers a visualization (TODO: verify)

### DIFF
--- a/contrib/swank-fancy-inspector.lisp
+++ b/contrib/swank-fancy-inspector.lisp
@@ -9,6 +9,15 @@
 (eval-when (:compile-toplevel :load-toplevel :execute)
   (swank-require :swank-util))
 
+(defun symbol-info (symbol)
+  (list (and (constantp symbol) (symbol-value symbol))
+	  (and (fboundp symbol) (symbol-function symbol))
+	  (and (boundp symbol) (symbol-value symbol))
+	  (and (find-class symbol nil) (find-class symbol))
+	  (find-package symbol)
+	  #+armedbear (get symbol 'sys::structure-definition)))
+
+    
 (defmethod emacs-inspect ((symbol symbol))
   (let ((package (symbol-package symbol)))
     (multiple-value-bind (_symbol status)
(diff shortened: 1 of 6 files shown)


easye:
@ioannad Can you help me review for SLIME style?

I can verify that all these changes work well with ABCL, (and other populate open implmentations SBCL, CCL, ECL).

@stassats Objects to https://github.com/slime/slime/pull/376/commits/d9092c3730049c6bb31e9514ef22cd71fb950928 as "not something he would like to see by default in his SLIME".  @ioannad How to accommodate both @stassats and @alanruttenberg here.

ioannad:
@easye: The style appears in accord with `CONRIBUTING.md`, apart from a couple of strange spaces, and apart from `contrib/swank-fancy-inspector.lisp`. Starting on line 107, the new function `lines-for-symbol-in-other-packages` is too large, 110x30, instead of 80x20 as required (`CONTRIBUTING.md` says "apart from that, we're not dogmatic"). 

On the other hand, quite a few methods and functions are very large in `contrib/swank-fancy-inspector.lisp`. Is there a reason for this? 

Additions to `slime.el`(l.3201,3203), `swank.lisp`(l.124), and `swank/backend.lisp`(l.165) also contain lines larger than 80 chars. 

Since this is the only place where `CONTRIBUTING.md` is "dogmatic" about, I could write a test to check this, and change the places where this happens.

About the objection from @stassats, I am not sure I understand what the disagreement is about, I couldn't find this discussion. To me it seems like these are not changes to the default: there is a universal variable that controls this behaviour (disabled by default), and the addition takes place in the fancy inspector, which I think is not the default.

easye:
@ioannad Thanks for the review.  @alanruttenberg and I will work on the raised issues.

contrib/swank-fancy-inspector.lisp
```diff
@@ -89,8 +98,44 @@
         ;; More package
         (if (find-package symbol)
             (label-value-line "It names the package" (find-package symbol)))
+
+	;; Same string
+	(lines-for-symbol-in-other-packages symbol)
+
```
easye: Conditionalize with UIOP?

easye:
Superseded by #500.

PR #378, opened 2017-04-06
Introduced *argument-stream*.

informatimago:
Introduced the *argument-stream* variable, bound while evaluating
expressions from slime, to the string-stream where the expression is
read from.  This allows the expression to further read from this
stream, to take additional arguments.

EXAMPLE:  Type in *slime-repl*:
              (* 2 (read swank:*argument-stream*))  21  RET
          --> 42

More typical example, using a symbol macro:

cl-user> (defun command-test ()
           (let ((token (read swank:*argument-stream*)))
             (print (list token token)))
           (values))
command-test
cl-user> (define-symbol-macro cmdt (command-test))
cmdt
cl-user> (command-test) 42

(42 42) 
; No value
cl-user> cmdt 42

(42 42) 
; No value
cl-user>

### DIFF
--- a/packages.lisp
+++ b/packages.lisp
@@ -191,4 +191,5 @@
            #:parse-string
            #:from-string
            #:to-string
-           #:*swank-debugger-condition*))
+           #:*swank-debugger-condition*
+           #:*argument-stream*))
--- a/swank.lisp
+++ b/swank.lisp
@@ -53,6 +53,18 @@
 The list has the form ((VAR . VALUE) ...).  Each variable VAR will be
 bound to the corresponding VALUE.")
 
+(defvar *argument-stream* nil
+  "This variable is bound while evaluating expressions from slime,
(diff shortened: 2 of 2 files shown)


informatimago:
Both changes can be accepted: symbols were added from different patches.

luismbo:
This seems pretty niche, which is both a pro and a con. :-D `*argument-stream*` seems like a weird name, that only makes sense for this specific use-case. How about `*eval-region-stream*` instead?

informatimago:
Whatever the name.  The functionality is useful.

stassats:
I don't like additional code burden for marginally useful features that few people will know about and even fewer will use it.

informatimago:
> I don't like additional code burden for marginally useful features that few people will know about and even fewer will use it.

What code burden? This is simply making a lexical variable global dynamic, and exporting the symbol!

PR #380, opened 2017-04-16
Add support for slime-eval-print-last-expression for Kawa scheme

aTick:
This PR implements eval-and-grab-output in contrib/swank-kawa.scm, allowing for slime-eval-print-last-expression to work when using slime with Kawa scheme.

Note that this implementation does not capture values printed to
standard output. It is not clear that dynamically rebinding the
standard output port is possible in Kawa.

### DIFF
--- a/contrib/swank-kawa.scm
+++ b/contrib/swank-kawa.scm
@@ -809,6 +809,12 @@
          (l (values-to-list (eval form env))))
     (apply cat (map pprint-to-string l))))
 
+(defslimefun eval-and-grab-output (env string)
+  (let ((form (read (open-input-string string))))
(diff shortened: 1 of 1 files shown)


luismbo:
Pushed your commit with a slightly tweaked title and with tabs replaced by spaces: f1a4cbce. Thanks!

aTick:
I let emacs do the indenting for me, and didn't pay attention to what whitespace it was using. I'll try to remember to get this configured correctly in the future.

Thanks for fixing that for me, and for the quick turnaround on this!

luismbo:
@aTick right, Emacs uses tabs by default, so that file should declare indent-tabs-mode: nil, probably.

PR #383, opened 2017-04-22
Issue#379 Remove bindings on SPC

psequeirag:
The explicit bindings on SPC are not strictly necessary in order to
display the arglist in the echo area thanks to the hook registered
with eldoc. Furthermore, these bindings may conflict with other modes
like parinfer, which relies on SPC being a self-insert-command to work
properly, and so these bindings are dropped.

## Notes
* I didn't succeed in my first attempt to run the tests with `make check`, so I haven't yet verified that I'm not breaking anything.
* I'm not sure if the slime-space could also be dropped because it's referenced in a line like this:
```(put 'slime-space 'delete-selection t) ; for delete-section-mode & CUA```

### DIFF
--- a/contrib/slime-autodoc.el
+++ b/contrib/slime-autodoc.el
@@ -172,8 +172,7 @@ If it's not in the cache, the cache will be updated asynchronously."
   "Toggle echo area display of Lisp objects at point."
   :keymap (let ((prefix (slime-autodoc--doc-map-prefix)))
 	    `((,(concat prefix "A") . slime-autodoc-manually)
-	      (,(concat prefix (kbd "C-A")) . slime-autodoc-manually)
-	      (,(kbd "SPC") . slime-autodoc-space)))
+	      (,(concat prefix (kbd "C-A")) . slime-autodoc-manually)))
   (set (make-local-variable 'eldoc-documentation-function) 'slime-autodoc)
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by luismbo** [comment]
With your changes, the documentation is no longer fontified (i.e., no pretty colors). (I'm using Emacs 25.1, FWIW.)

**Comment by psequeirag** [comment]
Hmm, I had been testing only with `slime-autodoc` enabled, and the arglist still shows fontified with it: 
slime-space-autodoc

If I don't load `slime-autodoc`, I don't even see an argslist at all in the echo area, even with an open connection to sbcl. I'm not sure if this is the same you see. Here are my settings in .emacs.d/init.el:
```elisp
(add-to-list 'load-path "~/projects/slime")
(require 'slime-autoloads)
(setq inferior-lisp-program "~/Downloads/sbcl/run-sbcl.sh")
(add-to-list 'slime-contribs 'slime-fancy)
```

psequeirag:
Also forgot to mention that I'm testing this with Emacs 24, I'll need to setup a 25 and retry.

In any case, the changes do disable displaying the arglist upon pressing SPC unless slime-autodoc is enabled. If you did not want that, then I'll have to try parinfer's approach of registering with post-command-hook.

ioannad:
Refers to Issue #379

PR #385, opened 2017-04-29
ecl improvements

dkochmanski:
General cleanup of ecl.lisp backend file.

Most notable changes:
- spawn is preferred communication style for threaded environment (nil otherwise)
- sentinel thread should work (register/find-thread are implemented)
- swank functions aren't shown in the backtrace
- order of interface implementation matches backend.lisp file
- implementation for `macroexpand-all` has been added
- added package local nicknames implementation

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -1316,6 +1316,7 @@ The default condition handler for timer functions (see
     "May the source be with you!"
     "Take this REPL, brother, and may it serve you well."
     "Lemonodor-fame is but a hack away!"
+    "Are we consing yet?"
     ,(format "%s, this could be the start of a beautiful program."
              (slime-user-first-name)))
   "Scientifically-proven optimal words of hackerish encouragement.")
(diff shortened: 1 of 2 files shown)


luismbo:
@dkochmanski looking good. Are you going to implement more interfaces? (Minor bikeshedding: maybe rename "backend: ecl:" and "swank: ecl backend:" to simply "ecl:".)

dkochmanski:
Thanks. Right now I'm adding local-package-nicknames extension to ECL, so this will be added too (protected by a feature-condition). If I won't run out of steam by then, I want to take a jab at implementing condition_variable_timedwait, so receive-if supports timeouts correctly.

Given that after that I'm still on a roll (unlikely), I'll dig into introspection interfaces.

dkochmanski:
OK, this is ready to be merged, unless you want me to change something.

dkochmanski:
I've added one encouragement sentence which I've heard on IRC and found it amusing, I hope you don't mind?

luismbo:
@dkochmanski I like it, yeah. :-) Thanks!

PR #386, opened 2017-04-30
fix "coding" parsing in emacs-style "-*-" lines

budden:
When encoding name in Emacs-style first line like ";; -\*- coding: utf-8; other: stuff -\*-" is immediately followed by a semicolon, SLIME failed to parse the name correctly. This is fixed by suggested change.


luismbo:
Thanks @budden.

PR #388, opened 2017-05-11
ecl: non-threaded build doesn't have mp:make-lock interface

dkochmanski:
Add feature check for non-threaded builds.

### DIFF
--- a/swank/ecl.lisp
+++ b/swank/ecl.lisp
@@ -1046,6 +1046,7 @@
 
 ;;;; Locks
 
+#+threads
 (defimplementation make-lock (&key name)
   (mp:make-lock :name name :recursive t))
 


luismbo:
@dkochmanski Will users get in trouble if they compile swank with threads then reload it with threads or vice-versa? (Just wondering if we could/should do something about that?)

luismbo:
Merged, thanks!

dkochmanski:
if ECL compiled with threads loads swank compiled with compiler not supporing threads (and vice-versa) it will fail miserably, but a good news is that it will be immediate. Regarding proper validation - I don't have good answer for that now, because ECL doesn't validate FAS files whatsoever.

One solution could be recompilation when: version or commit doesn't match / if the match, if `*features*` doesn't match. It's not perfect. If I come up with something, I'll let you know.

I think that it's pretty rare situation, when user compiles swank with one ECL version and loads it with another (but build from the same commit). I would just skip this problem until CDR talking about ABI is implemented, or ECL is smarter wrt resolving symbols.

dkochmanski:
there is still one problem with ecl build without threads - any condition normally invoking sldb breaks the connection. I don't have solution for that yet.

luismbo:
@dkochmanski I guess SLIME could codify features into the fasl directory's name, but I'm not sure how robust that would be. Anyway, this sort of issue is likely to affect other backends as well, FWIW.

attila-lendvai:
doesn't ASDF solve that problem, too? and if not, maybe it could/should be patched to include not only the arch, but some important features into the fasl dir's name?

dkochmanski:
I think that proper way to do that is implementing
https://common-lisp.net/project/cdr/document/10/index.html

Regarding ASDF - it just verifies commit, not different build variants.

PR #397, opened 2017-08-23
Do not perform arithmetic operations with NaN in SBCL

christophejunke:
Calling (= F F) or (> F ...) when F is NaN throws an arithmetic
exception in SBCL. Instead, we use the built-in predicate.

### DIFF
--- a/contrib/swank-fancy-inspector.lisp
+++ b/contrib/swank-fancy-inspector.lisp
@@ -936,12 +936,14 @@ SPECIAL-OPERATOR groups."
 
 (defmethod emacs-inspect ((f float))
   (cond
+    ((or #+sbcl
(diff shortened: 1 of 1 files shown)


luismbo:
Thanks!

christophejunke:
Thank you! Slime is great!

luismbo:
@christophejunke one potential improvement to this would be to implement a NaN predicate as a swank interface. Do you feel like taking a shot at that?

christophejunke:
Basically, one `definterface` in backend.lisp with a default body and one `defimplementation` for each implementation in `swank`? I'll try to do that.

luismbo:
Yes, exactly.

PR #398, opened 2017-08-24
Define an interface for detecting float NaN and Infinity

christophejunke:
Let implementation provide specialized functions for detecting special floating point values.

Maybe the portability layer could be implemented in IEEE-FLOATS[1], in which case Slime would have to  depend on that library. I don't know which way is the best.

[1] See issue https://github.com/marijnh/ieee-floats/issues/5


luismbo:
Great, thanks! Regarding ieee-floats, the current party line is that swank should have no external dependencies.

PR #403, opened 2017-09-30
wip: ecl: make async communication stable

dkochmanski:
Previous implementation is instable for threaded communication, because CV may block of of the blue. This version is more stable.

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -784,13 +784,11 @@ first."
   (create-server :port port :style style :dont-close dont-close))
 
 (defun accept-connections (socket style dont-close)
-  (let ((client (unwind-protect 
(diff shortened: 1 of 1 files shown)


dkochmanski:
still experience some glitches from time to time

luismbo:
We can probably change the default communication style to the previous method if you'd like more time to stabilize the multi-threading support. Would that help?

dkochmanski:
Hey, thanks. According to reports on ecl-devel mailing lisp, changing communication style to NIL doesn't fix the problem. Let me work on it some more, so I can pinpoint the issue and fix it.

luismbo:
Oh, OK. Good luck.

PR #405, opened 2017-10-19
Added slime description mode.

knobo:
For syntax hightlighting `*slime-description*` buffer


knobo:
The syntax highlighting does not work for anything but slime-describe-symbol. Got to make some improvments to that.

luismbo:
@kobo ACK. Keep us posted!

PR #412, opened 2017-11-18
Add support for :FD-HANDLER communication style to ECL backend

nickg:
Also tested with `NIL` and `:SPAWN` communication styles.

### DIFF
--- a/swank/ecl.lisp
+++ b/swank/ecl.lisp
@@ -120,13 +120,26 @@
 ;;;
 ;;;   remove-sigio-handlers (socket)
 
-;;; Call FN when Lisp is waiting for input and SOCKET is readable.
-;;;
-;;;   add-fd-handler (socket fn)
+#+serve-event
+(progn
+  (defvar *fd-handlers* nil)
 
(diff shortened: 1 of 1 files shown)


luismbo:
@dkochmanski any chance you could review this?

dkochmanski:
mmm, ECL has pink label :-)

PR looks fine to me at first glance, but SLIME having preferred communication set to `:fd-handler` doesn't start REPL which I could use on ECL 16.1.3 (tried only this version so far). I've removed FAS from ~/.slime directory and things got compiled anew, but it didn't help.

obraz

nickg:
@dkochmanski you need to run something like `(loop (serve-event:serve-all-events))` to actually get it to process the events. 

My use case is running a SWANK server inside a program with a single-threaded event loop that I then `slime-connect` to from Emacs. I'm not sure if there's a way to get Emacs to call `serve-all-events` itself when it starts the inferior-lisp?

luismbo:
@nickg perhaps you could add a note about this to https://common-lisp.net/project/slime/doc/html/Communication-style.html#Communication-style?

PR #421, opened 2017-12-25
Error while loading slime because of accent

nocaoper:
The accentuated i in LUIS, in line 3 causes a forced break in ubuntu 16 (bash mode).

image

after fixing the i it becomes normal:

image


luismbo:
Heh. So sad we can't spell my name properly. 🙃 Thanks!

nocaoper:
:D, thanks for replying, ha, this made my day!

PR #423, opened 2018-01-10
Several patches applied to the Debian package.

sebastien-villemot:


luismbo:
Applied as 5ec2b5c7, e17574cc and d0164b88 with minor tweaks to the commit message: the first line, like an e-mail subject, shouldn't end with a period. Thanks!

sebastien-villemot:
@luismbo Thanks!

PR #425, opened 2018-01-30
slime-version-string: make behavior consistent with documentation

sebastien-villemot:
Return nil if version cannot be parsed from slime.el, instead of crashing.

### DIFF
--- a/swank-loader.lisp
+++ b/swank-loader.lisp
@@ -141,12 +141,12 @@ operating system, and hardware architecture."
 Return nil if nothing appropriate is available."
   (with-open-file (s (merge-pathnames "slime.el" *source-directory*)
                      :if-does-not-exist nil)
-    (and s
-         (loop with prefix = ";; Version: "
-               for line = (read-line s nil :eof)
-               until (eq line :eof)
-               when (string-starts-with line prefix)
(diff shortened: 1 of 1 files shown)


swank-loader.lisp
```diff
@@ -141,11 +141,12 @@ operating system, and hardware architecture."
 Return nil if nothing appropriate is available."
   (with-open-file (s (merge-pathnames "slime.el" *source-directory*)
                      :if-does-not-exist nil)
-    (loop with prefix = ";; Version: "
-          for line = (read-line s nil :eof)
-          until (eq line :eof)
-          when (string-starts-with line prefix)
-            return (subseq line (length prefix)))))
+    (and s
```
luismbo: This is a bit of a bikeshed comment, but I think `when` would be more appropriate here. What do you think?

sebastien-villemot:
I chose `and` because it was used before commit 21ae13c619c1040228e2db88db08587de441d980 for that very purpose. But of course `when` is also fine, so feel free to pick the one you prefer.

luismbo:
@sebastien-villemot let's go with `when`.

luismbo:
Thanks!

PR #428, opened 2018-02-07
Call swank:inspector-call-nth-action with a single colon, not ::

ecraven:
`inspector-call-nth-action` is a normal slime function, it can be called with a single colon.
I'm working on a swank for R7RS Schemes, and kawa can not read :: (that is used for type annotations).

This should not influence CL swank at all.


ecraven:
> Merged #428.

Thank you!

PR #431, opened 2018-02-18
Small fix to common-lisp-hyperspec--read-format-character

J-Robertson:
The call to `completing-read` in this function appears to have the arguments in the wrong order:
The require-match argument given is `#'boundp`(which defaults to `t`), history is given as `nil` and default is the history variable. This actually has little effect on the default completing-read function which still works fine. When `completing-read-function` is different (eg. in helm or ivy), however, it may misbehave. I am unsure of the effects on helm users but this is useless with ivy.

### DIFF
--- a/lib/hyperspec.el
+++ b/lib/hyperspec.el
@@ -1237,8 +1237,8 @@ If you copy the HyperSpec to another location, customize the variable
 	char-at-point
       (completing-read
        "Look up format control character in Common Lisp HyperSpec: "
-       common-lisp-hyperspec--format-characters nil #'boundp
-       nil nil 'common-lisp-hyperspec-format-history))))
+       common-lisp-hyperspec--format-characters nil t nil
+       'common-lisp-hyperspec-format-history))))
 
 (defun common-lisp-hyperspec-format (character-name)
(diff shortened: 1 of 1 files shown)


luismbo:
@J-Robertson what does "misbehave" mean? Can you help me test this?

J-Robertson:
Right. Originally, when using ivy's completing read function, no completions were listed. After brief testing I found that the reason it broke was that the argument for a default value of the completion was the symbol `'common-lisp-hyperspec-format-history` rather than a string. This is how it misbehaves.

I have tested the change I made with the default completing-read function and with ivy's. Both worked as they should. Furthermore I made sure to match the arguments with the documentation of completing-read and made the call consistent with all other calls to completing-read in this file and the project.

Given that this was a minor rewording to fix an invocation of a function that was inconsistent with the rest of the project and with the documentation, I figured little further testing was necessary.

If you want I can do some more when I get to my computer in a few hours.

luismbo:
@J-Robertson thanks for the detailed description/rationale. I was having trouble what was misbehaving because helm seems to be fine with `'common-lisp-hyperspec-format-history` being the default value, despite it being non-sensical.

PR #446, opened 2018-04-04
Fix some warnings in Emacs 25.1.

Ferada:
Not sure how backwards compatible this is, but it fixes some compilation warnings on Emacs 25.1.

The font lock one should be the flush function I think? The rest are simply replacing deprecated aliases; the `DEFCUSTOM` call makes for a nicer menu.

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -281,7 +281,10 @@ argument."
   "List of functions to perform completion.
 Works like `completion-at-point-functions'.
 `slime--completion-at-point' uses this variable."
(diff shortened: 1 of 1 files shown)


stassats:
Many of these functions are missing on 24.5

Ferada:
Meaning it shouldn't be merged, yes? I don't have 24.5 available, would you be happy to just take the `DEFCUSTOM` part if I update it?

Issue #427, opened 2018-02-01
slime-make-tramp-file-name broken on Emacs 26


daewok:
I've noticed two issues with `slime-make-tramp-file-name` on Emacs 26.

First, `tramp-make-tramp-file-name` now takes 6 required arguments. This is rather easy to fix by just passing `nil` for the new arguments.

Second, passing `nil` as the method doesn't seem to work any more, even with `tramp-default-method` and co. set. I've confirmed that with the same config on Emacs 25, I can use slime-tramp to access files using ssh, but the same config and same remote computer on Emacs 26 (accounting for the new arguments) doesn't work unless I specifically put "ssh" as the method on the call to `tramp-make-tramp-file-name`; I just get an empty buffer and the error "Not a Tramp file name: \"/user@host:\"" when I C-x C-f in that buffer (using Helm). I'm not sure how to go about fixing this second error.

aaronbieber:
**Expectation:** I can open `helm-find-files` and type `/copernicus:/home/` to open the `home` directory on the network machine `copernicus` via SSH (my default protocol).

**Actual result:** A lisp error in `helm-find-files-get-candidates`, `Not a Tramp file name: "/copernicus:"`

I am encountering this same issue without Slime; simply opening `helm-find-files` and typing in a TRAMP name without protocol hits this error. Note that this procedure works perfectly in Emacs 25, but suddenly in Emacs 26 it does not.

The stack trace shows indicates that the issue resides within `helm-find-files-get-candidates` (see trace below). What is interesting is that the trace shows that `/ssh:` is getting prepended to my path in `helm-get-candidates`, but is removed within the failing function.

I can get around it by typing in `/ssh:copernicus:/` (which is annoying).

Here is the full stack trace: https://gist.github.com/aaronbieber/573268fe643c05dfb36ed9fba5725b0c

aaronbieber:
See the reply to the Helm issue here: https://github.com/emacs-helm/helm/issues/2090#issuecomment-418089720

This is possibly related to the Tramp configured syntax (`M-x tramp-change-syntax RET simplified RET`), where "simplified" means you can leave off the method.

daewok:
Thanks for the pointer! Turns out the method being required is stated right in the news file: `*** The method part of remote file names is mandatory now.`

I think I've got enough now to actually fix this issue.

daewok:
Closed by

PR #459, opened 2018-09-03
Update slime-make-tramp-file-name for emacs 26.

daewok:
Emacs 26 added two new required arguments to `tramp-make-tramp-file-name`. Additionally, the method part of remote file names is mandatory.


daewok:
Closes #427.

PR #468, opened 2018-10-29
Support uiop:define-package

knobo:
Support slime-find-package-definition-regexp with uiop:define-package


guicho271828:
:+1:

Ambrevar:
This patch looks good to me, can it get merged?

Issue #213, opened 2014-12-02
Restore the slime-presentation-streams contrib


luismbo:
It was disabled due to issue #212. https://sourceforge.net/p/sbcl/mailman/message/33102884/ outlines how to reimplement it properly.

luismbo:
@phmarek fixed some newer breakage here: https://github.com/phmarek/slime/commit/e3c3540786a000d7afee1cd420b99c3ba3c5f7f7

luismbo:
@snuglas how likely do you think it would be for SLIME to have some way to hook into pretty printing? Would it help to send an e-mail to sbcl-devel (or similar) describing slime-presentation-streams's needs?

scymtym:
Another data point for that discussion would be https://github.com/McCLIM/McCLIM/tree/master/Experimental/pprint.

I think it also tried to turn pieces of pretty print output into presentations.

CC @dkochmanski

snuglas:
probably best to ask on sbcl-devel, even better if you can also propose
some mechanism by which to achieve the outcome.

>

luismbo:
@phmarek, @scymtym: are any of you inclined to propose an API and/or patch?

scymtym:
> @scymtym: are any of you inclined to propose an API and/or patch?

I don't have enough time at the moment, sorry.

phmarek:
Well, why not take _stassats_ idea from the ML? Or does that solve a different problem?

luismbo:
@stassats has [your idea](https://sourceforge.net/p/sbcl/mailman/message/33102884/) matured in the past 4 years? :-)

stassats:
Yes, it's prime for somebody (read _not me_) to implement it and integrate it into Slime.

luismbo:
@stassats I wonder if it'd compose better to have make-pretty-stream be a generic function that dispatches on the target stream? That way we could add methods for SLIME streams and leave room for other prettiness hacks without setting a global variable.

stassats:
It's not too early for CLOS (i.e. when Slime is loaded). The target stream would have to be subclassed and methods are easily dynamically scoped.

luismbo:
BTW, I recently noticed that `swank-presentation-streams` has two mechanisms for hooking into object printing, one of which is only used when in `*use-dedicated-output-stream*` mode that is much less intrusive and doesn't require hooking into `make-pretty-stream`. I think this explains how @phmarek is using this contrib when I had the impression that the contrib didn't work at all.

[...]

PR #475, opened 2018-11-29
Fix print-unreadable-object for newer SBCL.

phmarek:
Partial fix for issue #213.


phmarek:
Why is Travis checking PR 456?? This is 475!

luismbo:
@phmarek D'oh, stassats's suggestion is painfully obvious in hindsight. Go for it.

PR #477, opened 2018-12-05
V2.22/abcl

easye:
December 2018 version of ABCL specific changes to SLIME.  

Unmerged as we (Alan Ruttenberg and Mark Evenson) refactored things push primitives into `swank-backend`.


easye:
Superseded by #500

PR #485, opened 2019-01-14
Fix race conditions in swank

spaghettisalat:
When swank kills threads, it checks whether they are alive before doing so. However, the threads can still die before kill-thread is actually called, leading to spurious errors like "Cannot interrupt the inactive process ...". The only reasonable way to prevent this is to ignore such errors. The issue is not only academic, with ECL I'm observing the behaviour regularly on slime startup.

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -583,7 +583,7 @@ recently established one."
              (when (and thread 
                         (thread-alive-p thread)
                         (not (eq thread (current-thread))))
-               (kill-thread thread))))
+               (ignore-errors (kill-thread thread)))))
           (t
            (warn "No server for ~s: ~s" key value)))))
 
@@ -1177,7 +1177,7 @@ event was found."
       (when (and thread
                  (thread-alive-p thread)
                  (not (equal (current-thread) thread)))
-        (kill-thread thread)))))
+        (ignore-errors (kill-thread thread))))))
 
 ;;;;;; Signal driven IO
 


stassats:
Why is kill-thread even being used? Can the slime threads be stoped gracefully?

spaghettisalat:
> other implementations of kill-thread don't signal errors when trying to kill dead threads, so maybe you can add ignore-errors to the ECL implementation only?

CCL and CMUCL do not, but at least SBCL does signal such errors. Wouldn't it be simpler to ignore errors for all implementations, instead of having to test all of them?

spaghettisalat:
> Why is kill-thread even being used? Can the slime threads be stopped gracefully?

I'm not familiar enough with the swank internals to answer this question or try to refactor the code as to not use kill-thread, sorry. I was just trying to fix this slightly annoying bug.

luismbo:
@spaghettisalat In general `ignore-errors` should be avoided since it can catch other errors (e.g., we could be passing the wrong object to kill-thread). On the other hand, the `kill-thread` swank interface does not currently define what should be done in exceptional situations and using ignore-errors within the implementation (as I had suggested) might have a negative effect on the `slime-list-threads`/`slime-thread-kill` functionality (e.g., if `kill-thread` fails we might want to know why). So, this is a step forward. In any case, in the future, we should remove this use of `kill-thread`. Thanks!

luismbo:
@spaghettisalat Having said that, as a step towards getting rid of `kill-thread` usage, it'd be useful to know exactly what conditions are triggering `kill-thread`. I don't see it being invoked at startup. What are the steps to reproduce this?

PR #486, opened 2019-01-20
add SWANK::RUN-HOOK-WITH-ARGS-UNTIL-SUCCESS

fiddlerwoaroof:
this makes the CL event dispatch work more like the ELisp dispatch and allows contribs to add events without requiring any changes to swank.lisp

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -104,6 +104,12 @@ include some arbitrary initial value like NIL."
   (dolist (function functions)
     (apply function arguments)))
 
+(defun run-hook-with-args-until-success (functions &rest arguments)
+  "Call each of FUNCTIONS with ARGUMENTS, stop if any function returns
+a truthy value"
+  (loop for hook in functions
+          thereis (apply hook arguments)))
+
 (defvar *new-connection-hook* '()
   "This hook is run each time a connection is established.
 The connection structure is given as the argument.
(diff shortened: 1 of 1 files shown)


fiddlerwoaroof:
As discussed on IRC:

```
<fiddlerwoaroof> It's also a bit surprising that I have to update swank.lisp 
    every time that I add a new event keyword
<fiddlerwoaroof> Shouldn't it do what the emacs function slime-dispatch-event 
    does and only fail if the contribs don't handle an unrecognized keyword?

... snip ...

<luis> fiddlerwoaroof: looking good. I'll see what we can do to avoid having to 
    change swank.lisp. It's unfortunate indeed.
<fiddlerwoaroof> luis: perhaps we could implement hooks?
<fiddlerwoaroof> i.e. just have a list of functions and call the next function 
    if a previous one returns a truthy value, just like the emacs lisp
    does?
<luis> fiddlerwoaroof: do you mean that each event should behave like a hook?
<fiddlerwoaroof> I was more thinking of using hooks to catch new events before 
    the dcase
<fiddlerwoaroof> So, swank does this:
    https://github.com/slime/slime/blob/master/swank.lisp#L1013
<fiddlerwoaroof> slime takes a different strategy, where the hooks have a 
    chance to prevent the dcase from seeing a new event
<fiddlerwoaroof> https://github.com/slime/slime/blob/master/slime.el#L2210
<luis> fiddlerwoaroof: so, let's add a  run-hook-with-args-until-success to 
    swank. Sound good.
```

fiddlerwoaroof:
Both those sound good, here's a commit that implements them

PR #489, opened 2019-01-25
Move target manipulation out of repl contrib

fiddlerwoaroof:
I'm not sure if this is exactly the right way to do this, but it moves the manipulation of targets entirely out of the repl contrib, without breaking the parts of slime and swank that depend on it already.


fiddlerwoaroof:
Great, has this been merged?

PR #495, opened 2019-02-06
Avoid trailing newlines in resend-form shortcut

Affonso-Gui:
Hello,

Regarding the `resend-form` shortcut command, in `slime-repl.el`.

This PR uses the optional `new-line` argument in `slime-repl-send-input` instead of directly inserting the newline, in order to avoid trailing newlines on consecutive `resend-form` evaluation.

Current situation is like the image below (comments were added afterwards evaluation):
slime-resend-form

### DIFF
--- a/contrib/slime-repl.el
+++ b/contrib/slime-repl.el
@@ -1415,8 +1415,7 @@ expansion will be added to the REPL's history.)"
   (:handler (lambda ()
               (interactive)
               (insert (car slime-repl-input-history))
-              (insert "\n")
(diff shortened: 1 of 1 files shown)


stassats:
That's because `slime-repl-history-trim-whitespaces` defaulted to NIL, I simply removed the variable, always trimming.

Affonso-Gui:
@stassats Thanks for your reply. I confirmed that the desired result is obtained by setting `slime-repl-history-trim-whitespaces`.

PR #496, opened 2019-02-08
Properly select next thread

fiddlerwoaroof:
We weren't keeping track of the right "next thread" correctly, so hitting `k` would cause the cursor in the thread list to jump two lines.  Fix that.


luismbo:
Fixed in https://github.com/slime/slime/commit/ec53bbd0023e7be3db5119c855d627ea5aa50570.

PR #497, opened 2019-02-10
ecl: frame-source-location: return error if frame source not found.

dkochmanski:


stassats:
I see that you don't have a commit bit, I sent you one, so that you can work on the ECL backend unimpeded.

PR #507, opened 2019-03-18
Fix unsatisfiable read-time conditional in swank-asdf contrib

scymtym:
The inner feature test in `#-asdf3 (eval-when … #+asdf3 t …)` is always false.


stassats:
What about
```(and #+asdf2 (asdf:version-satisfies (asdf:asdf-version) "2.14.6")```

scymtym:
Right. Updated.

PR #512, opened 2019-05-11
Added load path to autoloads.

Ergus:
This is needed in order to make slime work in emacs 27. Because the new package-quickstart.el that is generated automatically needs to know where to search for slime. 

Otherwise we need to set the slime directory by hand which is not consistent with use-packages.

### DIFF
--- a/slime-autoloads.el
+++ b/slime-autoloads.el
@@ -14,6 +14,9 @@
 
 ;;; Code:
 
+(add-to-list 'load-path (directory-file-name
+                         (or (file-name-directory #$) (car load-path))))
+
 (autoload 'slime "slime"
   "Start a Lisp subprocess and connect to its Swank server." t)
 


stassats:
Why is it adding the first item on load-path to load-path?

luismbo:
What's package-quickstart.el and what's use-packages? It seems very strange to have the autoloads add themselves to the load path.

Ergus:
> What's package-quickstart.el 

The Emacs startup will change a lot in the next release. There will be 2 configuration files: early-init.el and init.el. The (package-initialize) call in init.el will be not needed anymore (because it will be implicit) and the first file (early-init.el) is optional. Apart from that the autoload will be grouped  in a single file to avoid the multiple open/close files which is usually slower than reading a simple file. For that purpose the package.el will create/maintain/and update automatically the so called package-quickstart.el this file is updated only when a new package is installed or when calling the interactive command  package-quickstart-refresh. Such optimization reduces the load time specially in hard drives.

> and what's use-packages? 

It is a very popular package created and maintained by John Wiegley (one of the emacs core maintainers/developers) to simplifies the packages configurations with autoloads and other optimizations. It is expected to go to the base emacs core in a no so far future. 
[use-packages](https://github.com/jwiegley/use-package)

> It seems very strange to have the autoloads add themselves to the load path.

No it is not. Actually it is a recommended command when people compile the init file because in those cases the load path does not have to do a recursive search for dependencies and also knows where to look for stuff related with the package when loading in different directory. You can see that all the packages with an *-autoload.el file start with that line (if they are up to date). Usually the file is autogenerated and it is not provided, but for some reason slime provides it and don't have that line.

luismbo:
Thanks for the pointers. (I know use-package well; that plural threw me off.)

Looks like `update-directory-autoloads` indeed adds that [weird load-path snippet](https://github.com/emacs-mirror/emacs/blob/d2d4916046e31e46598f0a0edbc65e75b8cb4cc3/lisp/emacs-lisp/autoload.el#L375 
). Sadly, the [commit that introduced it](https://github.com/emacs-mirror/emacs/commit/4fac34cee97a3a5c3ccd86ecdd598317d86102b8) doesn't quite explain what's going on. Since `add-to-list` will not add duplicated elements, my best guess is that this is a convoluted way to write:

```elisp
(let ((current-directory (file-name-directory #$)))
  (when current-directory
    (add-to-list 'load-path (directory-file-name current-directory))))
```

We should probably generate the autoloads automatically, but until we do, then this looks like the way to go.

luismbo:
Merged, thanks @Ergus.

alexander-yakushev:
This commit breaks our setup, it adds `~/.emacs.d/` to load-path, and that prevents SLIME from loading.

luismbo:
@alexander-yakushev brilliant. 🤦‍♂️ Are you getting SLIME from MELPA or are you loading slime-autoloads manually?

alexander-yakushev:
I'm installing it via MELPA.

Ergus:
I need to ask:

1) Why it ads ~/.emacs.d/ instead of ~/.emacs.d/slime-versionnumber/ (this is not what the patch says)

2) Why such thing prevents slime from loading?

Any way I thing that the autoloads should be auto generated, not hardcoded. Any way I just tries emacs 26.1 and it works for me. Also installing form melpa.

luismbo:
Please let me know if f34e8ec fixes things (@alexander-yakushev) and doesn't break things (@Ergus). It should take a little bit to reach MELPA.

Ergus:
Hi: 

I still need more information, because it is working for me out of the box in emacs 26.2. Please it is important for me to reproduce the issue if still there, but I can't. If you still have it, please give more details about your setup.

luismbo:
@Ergus I've reproduced it as follows:

```;; file: test-home/.emacs.d/init.el
(require 'package)
(add-to-list 'package-archives
             '("melpa" . "http://melpa.milkbox.net/packages/") t)
(package-initialize)
(setq inferior-lisp-program "sbcl")
```

Then launched emacs with `HOME=test-home emacs`, installed slime (noticed that `slime-autoloads.el` was byte compiled into `slime-autoloads.elc`), then restart emacs. The autoloads won't be available. (FWIW, I didn't see `~/.emacs.d` being added to the load-path.)

alexander-yakushev:
The latest version from MELPA works, thank you!

I guess `~/.emacs.d` because of `(directory-file-name (car load-path))`, and that depends on what is the first item on load-path prior to that.

luismbo:
@alexander-yakushev OK, that can happen if `(car load-path)` is not a directory or is otherwise transformed by `directory-file-name`.

rprimus:
Wed Jun 26 16:46:36 BST 2019

The easiest way to solve this is to remove `slime-autoloads.el` from the repo and let `package-generate-autoloads` generate the `correct` autoloads file.

To test this yourselves, in your ELPA directory:
1. rename slime-autoloads.el
2. eval: (package-generate-autoloads “slime” “the ELPA slime directory”

Reference:
https://github.com/joaotavora/sly-quicklisp/issues/3

luismbo:
@rprimus yes, that makes sense. However, that would break the use case of grabbing SLIME via `git pull` and be instantly ready to run it without a `make` step. I'm not suggesting we should support that use case at all costs. Indeed, having to maintain a list of hotspots that are byte-compiled explicitly by SLIME is a bit too much IMHO. But, right now, the status quo seems to work well enough.

rprimus:
Wed Jun 26 17:36:36 BST 2019

@luismbo 

A solution to keep things in sync would be to replace the current `slime-autoloads.el` with one generated by `package-generate-autoloads`.

This would ensure both type os use-cases will be using the same file.

PR #521, opened 2019-08-18
Fix SWANK/SBCL:COMPILER-NOTE-LOCATION for non-file-backed streams

scymtym:
Calling `pathname` on an arbitrary stream can signal an error.

### DIFF
--- a/swank/sbcl.lisp
+++ b/swank/sbcl.lisp
@@ -558,8 +558,13 @@ information."
             (sb-c::compiler-error-context-original-source context)))
           ((typep condition 'reader-error)
            (let* ((stream (stream-error-stream condition))
-                  (file   (pathname stream)))
-             (unless (open-stream-p stream)
+                  ;; If STREAM is, for example, a STRING-INPUT-STREAM,
+                  ;; an error will be signaled since PATHNAME only
+                  ;; accepts a "stream associated with a file" which
+                  ;; is a complicated predicate and hard to test
+                  ;; portably.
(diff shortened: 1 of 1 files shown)


stassats:
Would checking for FILE-STREAM be enough to avoid ignore-errors?

scymtym:
> Would checking for FILE-STREAM be enough to avoid ignore-errors?

My initial thought as well, but apparently not: [pathname](http://www.lispworks.com/documentation/HyperSpec/Body/f_pn.htm) says the *pathspec* is a [pathname designator](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_p.htm#pathname_designator) which in case of a stream means a [stream associated with a file](http://www.lispworks.com/documentation/HyperSpec/Body/26_glo_s.htm#stream_associated_with_a_file) which includes synonym, two-way, broadcast, etc. streams.

stassats:
But in that case, can it be any of these? And how do you get slime to process compiler notes for string-streams?

scymtym:
> But in that case, can it be any of these? And how do you get slime to process compiler notes for string-streams?

One way this can happen is reading from a string-stream at load-time (or generally in the context of `swank:call-with-compilation-hooks`).

Putting this into any file of a system
```lisp
(read-from-string "#o8")
```
and loading the system using `M-x slime-load-system` reproduces the problem.

stassats:
Ok, I don't have time to look for a better solution, merged.

PR #526, opened 2019-09-01
Uiop define package

knobo:


knobo:
Maybe I should have reopened the previous pull-request #468. I don't know if that is possible.
How about moving the `:`. What do you think?
``` lisp
  "^(\\(cl\\|common-lisp\\|uiop\\|uiop/package\\)?\\(:?defpackage\\|define-package\\)\\>[ \t']*")
```

luismbo:
@knobo Your suggestion to move `:` would match things like `(:defpackage ...)`. Probably not a big issue, but it's weird. Anyway, thanks.

PR #531, opened 2019-09-20
Added line, column kw args to swank-compile-string

drmeister:
This will pass the 1-based line and column source position
info (or NIL) to swank-compile-string.
The line and column source position information is
where the string to be compiled starts in the file
with the given filename.

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -2496,9 +2496,9 @@ Record compiler notes signalled as `compiler-condition's."
   "Compile STRING (exerpted from BUFFER at POSITION).
 Record compiler notes signalled as `compiler-condition's."
   (let* ((offset (cadr (assoc :position position)))
-         (line-column (assoc :line position))
-         (line (cadr line-column))
-         (column (caddr line-column)))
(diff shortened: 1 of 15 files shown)


swank.lisp
```diff
@@ -2495,7 +2495,10 @@ Record compiler notes signalled as `compiler-condition's."
 (defslimefun compile-string-for-emacs (string buffer position filename policy)
   "Compile STRING (exerpted from BUFFER at POSITION).
 Record compiler notes signalled as `compiler-condition's."
-  (let ((offset (cadr (assoc :position position))))
+  (let* ((offset (cadr (assoc :position position)))
+         (line-column (assoc :line position))
+         (line (cadr line-column))
+         (column (caddr line-column)))
```
luismbo: Maybe turn this into:

```
   (line-column (cdr (assoc :line position)))
   (line (first line-column))
   (column (second line-column)))
```

if you agree that's more readable. Not a big deal.

drmeister: That looks better - I'll follow your suggestion.

PR #536, opened 2019-11-11
abcl: consolidate fixes and features

easye:
(Alan Ruttenberg)(Mark Evenson)

Consolidated fixes and features to working across the most common ABCL
releases, namely development (1.6.0-dev), current (1.5.0) and
historical (1.4.0).  A majority of the new features depend on the
ABCL-INTROSPECT tooling introduced with abcl-1.5.0, to which we
intend to conditionalize their use on a backwards compatible strategy.

From <https://github.com/slime/slime/pull/477/commits/c7c72d7b0f5cd6473af3488202ea1eca24d58202>.

Enable goto source location to find definitions in JAR archives.

For primitives compiled from Java source, return implementation source
location designated by the logical pathname associated with 'SYS:SRC'.

Finding functions and methods works well, other than when methods are
defined as part of a DEFGENERIC form, or implicitly as with :report on
condition defining PRINT-OBJECT.

FIND-DEFINITIONS include symbols and specials defined in Java.

Redirect source lookup to system jar if it's not where it claims to be.

Show stack trace when inspecting Java conditions.

Implement an inspector for Java classes.

Print internal variables in Java stack traces.

Prettier function name if we can glean information about otherwise
anonymous function.

Fix so that edit slot action in CLOS object inspector works.

Don't use the ABCL native printObject() for frames, so we can get prettier
function display.

Compile DEFIMPLEMENTATION forms so that they are visible to edit definition.

Fix source-location (used by view frame source) to work in general and
in particular for local function - go to the source of their owner.

Don't show bogus locals for Java stack frames.

Enables load of xref package.

Only add ABCL-INTROSPECT once to *FEATURES*.

Robustly find intended class of Java objects.

Tweak to invoke hyperspec in inspector (offered in symbol browser if a
CL symbol) - *SLIME-INSPECTOR-HYPERSPEC-IN-BROWSER* if t shows
hyperspec in emacs buffer despite setting of
browse-url-browser-function.  This should eventually be moved into
more generalized Swank facilities if useful to others.

### DIFF
--- a/swank/abcl.lisp
+++ b/swank/abcl.lisp
@@ -3,13 +3,30 @@
 ;;; swank-abcl.lisp --- Armedbear CL specific code for SLIME.
 ;;;
 ;;; Adapted from swank-acl.lisp, Andras Simon, 2004
+;;; New work by Alan Ruttenberg, 2016-7
 ;;;
 ;;; This code has been placed in the Public Domain.  All warranties
 ;;; are disclaimed.
 ;;;
 
 (defpackage swank/abcl
-  (:use cl swank/backend))
+  (:use cl swank/backend)
+  (:import-from :java
+                #:jcall #:jstatic
+                #:jmethod
+                #:jfield 
+                #:jconstructor
+                #:jnew-array #:jarray-length #:jarray-ref #:jnew-array-from-array
+                #:jclass #:jnew #:java-object 
+                ;; be conservative and add any import java functions only for later lisps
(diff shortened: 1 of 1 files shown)


easye:
As requested by Stas and Luis, a single merge that just changes the ABCL implementation resulting in a vastly more useful SLIME experience on Java.  Best used with abcl-1.5.0 or the (unreleased) abcl-1.6.0.

PR #537, opened 2019-11-13
Fix swank.asd for implementations which don't allow package modificat…

easye:
…ions

It is unfortunately implementation dependent as to whether evaluating
DEFPACKAGE forms refering to an existing package modify the existing
package or not.  The overloading of the SWANK-LOADER package to
contain both ASDF definitions as well as exporting external symbols
breaks the use of swank.asd under implementations such as ABCL.

### DIFF
--- a/swank.asd
+++ b/swank.asd
@@ -19,10 +19,10 @@
 ;; This code has been placed in the Public Domain.  All warranties
 ;; are disclaimed.
 
-(defpackage :swank-loader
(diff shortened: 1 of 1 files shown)


stassats:
Does it have to define a package at all?

easye:
> Does it have to define a package at all?

Indeed it doesn't. Please see <https://github.com/slime/slime/pull/538>.

PR #540, opened 2019-12-19
fix spurious detection of allegro modern-mode

gendl:
Hi,

The `(eql 'h 'H)` test for Allegro MoDeRn-mode was incorrectly detecting modern-mode when evaluated in an ANSI mode image with `(readtable-case *readtable*)` set to `:invert`. 

This changes the test to check directly for whether a built-in CL symbol is in uppercase or not, which should be more reliable, because modern-mode images differ from ANSI mode images in precisely this regard. 

Thanks,

 Dave

### DIFF
--- a/swank-loader.lisp
+++ b/swank-loader.lisp
@@ -97,7 +97,7 @@
   #+lispworks (lisp-implementation-version)
   #+allegro   (format nil "~@{~a~}"
                       excl::*common-lisp-version-number*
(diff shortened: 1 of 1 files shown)


stassats:
Isn't SYMBOL-NAME redundant?

gendl:
Indeed, `symbol-name` can be jettisoned because a symbol is already a string designator.  Just pushed a new commit w/o `symbol-name`.

PR #542, opened 2020-01-15
Ignore spawn errors when warming up cache

tdrhq:
In LispWorks when running a build script with `-build`, multiprocessing is not available by default. This makes it very convoluted to load swank (something like `(mp:initialize-multiprocessing :swank-loader nil (intern "INIT" "SWANK-LOADER") :load-contribs t)`).

That's unnecessary, and this pull request should fix it. (In particular, I had a weird situation where I *couldn't* launch multiprocessing during the build process.)

### DIFF
--- a/contrib/swank-asdf.lisp
+++ b/contrib/swank-asdf.lisp
@@ -510,9 +510,12 @@ already knows."
 ;; since it accesses a file-system, so run it once at the background
 ;; to initialize caches.
 (when (eql *communication-style* :spawn)
-  (spawn (lambda ()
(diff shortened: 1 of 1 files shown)


luismbo:
I think swank should itself enable multi-processing earlier when communication style is spawn.

stassats:
With SSDs, is that form even needed?

luismbo:
@stassats good point. Removed it. @tdrhq this should fix your problem. Thanks.

tdrhq:
Thanks @luismbo & @stassats!

PR #545, opened 2020-02-07
Fixes for ABCL

easye:
1.  Fix finding files within JAR archives 

2.  Fix inspector for Java methods and fields


stassats:
You can commit ABCL fixes yourself.

PR #547, opened 2020-02-13
sbcl: Filter out invalid definition sources in find-definitions

xach:
They can't be visited, so filtering them eliminates this error:

    Error: DEFINITION-SOURCE of blah did not contain meaningful
    information

This commonly arises when a method is defined without a corresponding
defgeneric.

### DIFF
--- a/swank/sbcl.lisp
+++ b/swank/sbcl.lisp
@@ -804,9 +804,22 @@ QUALITIES is an alist with (quality . value)"
          name
          (sb-introspect::definition-source-description source-location)))
 
+(defun categorize-definition-source (definition-source)
+  (with-definition-source (pathname form-path character-offset plist)
+    definition-source
+    (let ((file-p (and pathname (probe-file pathname)
+                       (or form-path character-offset))))
+      (cond ((and (getf plist :emacs-buffer) file-p) :buffer-and-file)
+            ((getf plist :emacs-buffer) :buffer)
+            (file-p :file)
+            (pathname :file-without-position)
+            (t :invalid)))))
+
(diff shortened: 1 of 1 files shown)


luismbo:
Thanks Xach.

stassats:
Isn't it useful to know that there are more methods defined, even if their location is unknown?

joaotavora:
FWIW, I agree with @stassats, I'm not cherry picking this one to SLY.

stassats:
What about ignoring only missing defgenerics?

luismbo:
@stassats good point. Fixed in 4dabbe24.

luismbo:
@stassats what's the point of fce533e4?

rigettizach:
The main pain point for me was getting frequent errors. I have nothing against seeing more things if it does not result in an error to pass over them.

luismbo:
@rigettizach is it because you debug-on-error turned on?

xach:
No, it's just because it's annoying to see it in the minibuffer - as though
something is going wrong in the system, when it's not.

It's especially annoying when there are two entries, and one cannot be
visited, because if there was only one, it would instantly go to the one
that *can* be visited.

luismbo:
ACK. You should be happy with 4dabbe2 then.

stassats:
> @stassats what's the point of [fce533e](https://github.com/slime/slime/commit/fce533e4d093a049ba8345e81be7bb33d5c1c70e)?

To use the macro when it's defined.

luismbo:
@stassats oh right. Need to test before merging next time. Or fix the CI. 🤦‍♂️ Thanks.

PR #551, opened 2020-03-14
Support for `*error-output*` and `*trace-output*` in `swank:eval-and-grab-output`

akater:
It is often desirable to get output from `time` and `trace` in Emacs, particularly in org-babel. I have a patch for org-babel ready that introduces the feature but I have not submitted it to Org yet, as it relies on the following patch to `swank` that extends `swank:eval-and-grab-output` to support emitting `*trace-output*`; the patch also introduces `*error-output*`, for good measure.

I did not find any mechanism already present in `swank` that would currently allow to grab outputs other than `*standard-output*`.

This change, as is, should not affect anything but it would be much better if we returned plist or alist instead which would become an interface-breaking change.

Note: when I (accidentally) tried to bind `*trace-output*` _after_ binding `*standard-output*`, I did not get any trace in the return value. I don't understand why this would happen, which is alarming. However, I've been using the suggested implementation with no issues for a couple of days, without signs of trouble.


luismbo:
The explanation for the behaviour you mention is probably because `*trace-output*` and `*standard-output*` point to the same stream. That also illustrates why this change is not acceptable. It'll make `slime-eval-print` and s`lime-eval-save` silently suppress output to some streams.

For your use case, I think you can simply wrap whatever form you're evaluating with the appropriate code for collecting the output, right?

akater:
@luismbo 

> because *trace-output* and *standard-output* point to the same stream

This is very confusing. Could you elaborate?

> I think you can simply wrap whatever form you're evaluating with the appropriate code for collecting the output, right?

Doesn't look like it. Suppose I wrap the form to return the following list:
`(list-of-original-values string-with-trace-output)`. This value will be piped to `swank:eval-and-glab-output`. The latter will format it, messing up the layout of values. Besides, why would standard output be collected in a way that differs so much from collecting trace output? This certainly feels wrong.

akater:
I just performed the following experiment:

- reverted my change, restarted Emacs
- made new buffer with lisp mode enabled
- typed “(time 0)”
- `M-x` `slime-eval-print-last-expression` which employs `slime-eval-save`

I did not get any trace output so I don't see how these streams are connected, and I don't see how `slime-eval-print` would suppress anything due to my patch. In fact, my patch would actually simply break `slime-eval-print` which I did not notice (I don't use destructucing-bind) but I'll fix this and corresponding bit in `slime-eval-save` soon, not altering their current behaviour.

akater:
I guess I see now: you are dissatisfied with the fact that this update would make some output neither displayed in a repl buffer, nor
- emitted into current buffer, in case of `slime-eval-print`
- saved into kill ring, in case of `slime-eval-save` 
?

This is indeed undesirable. There is a straightforward fix to this: make slime-repl-eval and slime-repl-print emit everything they have received. However, it's not clear what would be the proper resolution.

Personally, I'm not fond of mixing output and values in the first place (more on this below) but this is certainly better than dropping some output altogether.

Nevertheless, separating output in such a manner that standard output does get emitted by `slime-repl-eval` and `slime-repl-print` while trace and errors do not, is arbitrary, as is letting output go together with values. Oftentimes, the output of time or trace is literally the sole purpose of the evaluation.

Quick fix is to make `slime-eval-print` and `slime-eval-save` emit
- errors
- trace
- standard output
- values

I believe this order is the most reasonable one. For a sample demo, calling `slime-eval-print-last-expression` on `(progn (print 'entry-point) (ignore-errors (/ 1 0)) (time 0))` in a lisp-mode buffer would give the following:

```common-lisp
(progn (print 'entry-point) (ignore-errors (/ 1 0)) (time 0))
; in: PROGN (PRINT 'ENTRY-POINT)
;     (/ 1 0)
; 
; caught STYLE-WARNING:
;   Lisp error during constant folding:
;   arithmetic error DIVISION-BY-ZERO signalled
;   Operation was (/ 1 0).
; 
; compilation unit finished
;   caught 1 STYLE-WARNING condition
Evaluation took:
  0.000 seconds of real time
  0.000004 seconds of total run time (0.000004 user, 0.000000 system)
  100.00% CPU
  651 processor cycles
  0 bytes consed
  

ENTRY-POINT 0
```

Org-babel could provide much better handling of this by putting different outputs into different blocks. I think this is relevant to note, as Common Lisp traditionally provides rich tools for its users which is arguably one of its characteristic qualities. While SLIME is one of such tools, evaluating in a plain text buffer enabled with a single major mode and cramming everything together below (even if it's just standard output) is not an example of rich tooling but of a kludge (that could e.g. break paren pairing mode in said buffer) and this it should not influence design decisions or evaluations thereof.

It's likely worth it to provide an abstraction for this repeated layout in `slime-repl-print` and `slime-repl-eval`, to ensure uniformity.

That said, another possible resolution would be for `slime-repl-eval` and `slime-repl-print` to only emit values and leave all other output to REPL buffer. To get org-babel working with outputs properly then, I'll have to either introduce a new interface, different from `swank:eval-and-grab-output`, or make `swank:eval-and-grab-output` configurable when it comes to handling streams.

What do you think?

luismbo:
@akater yeah, making `swank:eval-and-grab-output` optionally grab more streams seems like a good idea. I don't have a clear view of how users use `slime-eval-print-last-expression`, so I'm leaning towards being conservative.

BTW, given `(write-line "1" t) (write-line "2" *error-output*) (write-line "3" t)`, you want the grabbed output to be "1 2 3" rather than "1 3 2", right?

luismbo:
>> because *trace-output* and *standard-output* point to the same stream
> This is very confusing. Could you elaborate?

Well, `*trace-output*` and `*standard-output*` are variables and you can bind them to the same object. (You could also bind them to different synonym streams that redirect to the same underlying stream, which is a bit more confusing.)

akater:
> Well, *trace-output* and *standard-output* are variables and you can bind them to the same object.

My confusion is due to the fact that neither standard output pops up in traces nor vice versa; how could any of these point to the same stream then?

Concerning the patch: my branch `grab-multiple-outputs` now makes `slime-eval-print` and `slime-eval-save` emit all output they recieve. This is only incompatible with the current behaviour in the sense that they sometimes will return more than they used to, which would mostly be warnings, given that users were unlikely to use these functions for expressions with `trace` expressions in them.

I understand and respect SLIME's conservatism and I believe this would be a useful addition (modulo this confusing dependency on order of parameters that we've been discussing). It would only introduce drastic incompatibilities for users who rely on the following functions to only emit standard output and values:
- `slime-interactive-eval`
- `slime-eval-print-last-expression`

Given that standard output is already generally unparseable, I don't think this would  break any scripts that try to parse the emitted strings, as such scripts are unlikely to exist and even if they do, they better employ some other mechanism to get the data. Other uses are non-critical.

This change will make `slime-eval-print-last-expression` closer to `slime-pprint-eval-last-expression` which does emit trace output already.

Directing error output to some other buffer, not in general being displayed, while evaluating in current buffer is not really unlike silently eating error output in the first place.

One particular application that I need this for is sane return in org-mode blocks. Not having this is very inconvenient: it basically forces org-based evaluations to stay less featureful than their ugly hackish equivalents aimed at `lisp-mode` buffers. Emitting `*standard-output*` into lisp-mode buffers makes zero sense already. `org-mode` buffers can actually process and contain different kinds of outputs (I do it already, in limited form, and rely on this daily). If you're against this change, please advise how should I emit trace (and error) output, cleanly separated from each other and from other output, in buffers where it is actually appropriate to emit to, like org buffers. If not, let's figure out the parameter-ordering-dependence issue and get this feature to work. Maybe somebody else familiar with SLIME could do this if you're not interested.

luismbo:
> My confusion is due to the fact that neither standard output pops up in traces nor vice versa; how could any of these point to the same stream then?

I don't understand the question. What does it mean for standard output to pop up in traces?

> This is only incompatible with the current behaviour in the sense that they sometimes will return more than they used to

Hmm. `slime-interactive-eval` only captures the return value, it doesn't grab any streams, does it? E.g. try `M-x slime-interactive-eval`, `(progn (write-line "hello") 'bye)`. You'll get BYE in the mini buffer and `hello` in the REPL (usually).

> Emitting *standard-output* into lisp-mode buffers makes zero sense already. 
 
It makes sense for scratch buffers, namely `M-x slime-scratch`.

akater:
>> My confusion is due to the fact that neither standard output pops up
>in traces nor vice versa; how could any of these point to the same
>stream then?
>
>I don't understand the question. What does it mean for standard output
>to pop up in traces?

If `*trace-output*` and `*standard-output*` pointed to the same stream, it seems to me that invoking `eval-and-grab-output`, as it is now, would either capture `(time 0)` output, together with any standard output, or would cause `(format t "wow")` output to appear in REPL where trace output goes. But none of this happens.

>> This is only incompatible with the current behaviour in the sense
>that they sometimes will return more than they used to
>
>Hmm. `slime-interactive-eval` only captures the return value, it
>doesn't grab any streams, does it? E.g. try `M-x
>slime-interactive-eval`, `(progn (write-line "hello") 'bye)`. You'll
>get BYE in the mini buffer and `hello` in the REPL (usually).

`slime-interactive-eval` calls `slime-eval-save` and `slime-eval-print` when called with certain prefix arguments. They both use `eval-and-grab-output`.

>> Emitting *standard-output* into lisp-mode buffers makes zero sense
>already. 
> 
>It makes sense for scratch buffers, namely `M-x slime-scratch`.

Run `slime`, open `*slime-scratch*`, eval `(format t "|")`. If you use `paredit`, it will break globally in the buffer. That's what I mean. Arbitrary output only makes sense in cleanly isolated areas, in buffers with multiple major modes. (At the moment, `org` does not shield from this either but it's a known issue and hopefully will be fixed.)

luismbo:
> If `*trace-output*` and `*standard-output*` pointed to the same stream, it seems to me that invoking `eval-and-grab-output`, as it is now, would either capture `(time 0)` output, together with any standard output, or would cause `(format t "wow")` output to appear in REPL where trace output goes. But none of this happens.

Oh, OK. You're right. My explanation indeed doesn't explain what you're describing. Can you try and reduce your example so we can understand what's going on?


> > This is only incompatible with the current behaviour in the sense that they sometimes will return more than they used to Hmm. `slime-interactive-eval` only captures the return value, it doesn't grab any streams, does it? E.g. try `M-x slime-interactive-eval`, `(progn (write-line "hello") 'bye)`. You'll get BYE in the mini buffer and `hello` in the REPL (usually).
> `slime-interactive-eval` calls `slime-eval-save` and `slime-eval-print` when called with certain prefix arguments. They both use `eval-and-grab-output`.

So, I'm happy to accept a patch that optionally collects the output of more streams. But I'd like to keep the current default behavior. In part out of conservativeness, but also because it kind of makes sense to me. For instance, in a UNIX shell, `foo > bar` also only redirects the standard output. It makes sense to me that, by default, compilation warnings would go the repl (or a dedicated buffer or whatever), debug messages would keep going to a repl, trace output to trace buffer, etc.

Your use case is also perfectly legitimate, so I think an option is the way to go.

But, I think it's weird to collect each stream independently, it seems better to collect everything into a single string output stream. What do you think?

> > Emitting *standard-output* into lisp-mode buffers makes zero sense already. It makes sense for scratch buffers, namely `M-x slime-scratch`.
> Run `slime`, open `*slime-scratch*`, eval `(format t "|")`. If you use `paredit`, it will break globally in the buffer. That's what I mean. Arbitrary output only makes sense in cleanly isolated areas, in buffers with multiple major modes. (At the moment, `org` does not shield from this either but it's a known issue and hopefully will be fixed.)

Indeed, it doesn't play well with paredit. An option there might be to emit the output as comments.

akater:
Thank you @luismbo. Looks like I have achieved complete compatibility now when it comes to internal SLIME behaviour.

However, external interface does change: I return alist with various outputs labelled explicitly. Alternatives to this are:

- return plain list and refer to individual elements by their positions

This is quite ugly. Right now, org-mode accesses different values and standard output with car and cadr when it filters by them (this limited filtering feature has been present in org-mode for some time).

- squash all outputs into a single string

I'm strongly aginst this. Outputs of different kinds are often differently structured. For example, it makes sense to turn on lisp-mode font-lock in trace-output chunks, as presented by SBCL at this moment. With font-lock on, traces look much better. Traces are in general fairly predictably structured, unlike standard-output. Error outputs are commented out, at least in SBCL. Overall, I find it useful to be able to filter outputs of different kinds and treat them differently. My approach to this is inspired by an advanced but niche and proprietary interface which might explain why this was not suggested earlier. Squashing all outputs together would, in my opinion, go strongly against the whole premise of having them separated. It has as much sense as squashing together output and values which SLIME currently does not do.

Nevertheless, introducing alist would break compatibility with org-mode, at least. I have a patch for org-mode that makes everything work as if nothing happened (with some new features available too) but I need to talk to orgmode developers and first of all, make sure SLIME is ok with `eval-and-grab-output` returning alists af the form:
```lisp
((VALUES . "VALUE") (*STANDARD-OUTPUT* . "some output") (*ERROR-OUTPUT* . "") (*TRACE-OUTPUT* . ""))
```

Meanwhile, `slime-eval-print` and `slime-eval-save` by default emit output and values returned as they used to. They have the following headers:
```lisp
(cl-defun slime-eval-print (string &optional (targets-to-print '(common-lisp:*standard-output* common-lisp:values))) ..)
```
which I hope is self-explanatory.

luismbo:
>     * squash all outputs into a single string
> 
> 
> I'm strongly aginst this. Outputs of different kinds are often differently structured. For example, it makes sense to turn on lisp-mode font-lock in trace-output chunks, as presented by SBCL at this moment. With font-lock on, traces look much better. Traces are in general fairly predictably structured, unlike standard-output. Error outputs are commented out, at least in SBCL. Overall, I find it useful to be able to filter outputs of different kinds and treat them differently. My approach to this is inspired by an advanced but niche and proprietary interface which might explain why this was not suggested earlier. Squashing all outputs together would, in my opinion, go strongly against the whole premise of having them separated. It has as much sense as squashing together output and values which SLIME currently does not do.

Font locking different stream outputs different makes perfect sense. But what about the time information you lose? Consider:

```
(write-line "entering FOO" *trace-output*)
(write-line "strange thing happened right about here" *debug-io*)
(write-line "returning from FOO" *trace-output*)
```

Isn't it useful to have these outputs interleaved rather than segregated?

I guess we could tag each output chunk/line with the stream that generated it. Do you think that'd be worth the trouble? Anyway, if you're happy with segregation for your use case, that's fine. I'm afraid github doesn't let me reopen this pull request, you'll have to open a new one.

PR #553, opened 2020-03-22
A few small fixes

phmarek:
Thanks for considering!

### DIFF
--- a/swank/rpc.lisp
+++ b/swank/rpc.lisp
@@ -8,7 +8,7 @@
 ;;; are disclaimed.
 ;;;
 
-(in-package swank/rpc)
+(in-package :swank/rpc)
 
 
 ;;;;; Input
--- a/swank/rpc.lisp
+++ b/swank/rpc.lisp
@@ -106,7 +106,7 @@
     (write-sequence octets stream)
     (finish-output stream)))
 
-;; FIXME: for now just tell emacs that we and an encoding problem.
+;; FIXME: for now just tell emacs that we had an encoding problem.
 (defun encoding-error (condition string)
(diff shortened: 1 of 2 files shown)


stassats:
When is fuzzy-completions called with non-simple strings?

phmarek:
I'm currently adding some other communication protocols; eg. `yason` returns strings with a fill-pointer.

stassats:
Then shouldn't it be doing all the appropriate coercions then?

phmarek:
Well, wouldn't that mean most of that translation effort would be wasted? Most functions just accept non-simple strings already.

Yes, of course I could put a special case for that one call in... but that's really awful, code-wise.

stassats:
As it stands, it's equally as awful to make a change for code that doesn't exist.

swank/rpc.lisp
```diff
@@ -8,7 +8,7 @@
 ;;; are disclaimed.
 ;;;
 
-(in-package swank/rpc)
+(in-package :swank/rpc)
```
luismbo: Helmut actually meant to write it like that. If we change it here, we should then change it in a bunch of other places.

phmarek: > As it stands, it's equally as awful to make a change for code that doesn't exist.

Well, this is a single place that doesn't accept a `string` but has more narrow restrictions.
I think ensuring that type within the same library isn't wrong.


> Helmut actually meant to write it like that. If we change it here, we should then change it in a bunch of other places.

I just noticed that even within swank it's not one-or-another. Should I amend my PR to change all over to keywords, as it seems to be custom (data are my QL libraries)?

PR #562, opened 2020-05-28
Profiling with metering for clasp

kpoeck:
* test with latest clasp
* tested all entries in the menu
* make check shows no changes


luismbo:
Thanks @kpoeck.

PR #563, opened 2020-06-06
abcl: guard against repeated invocation to SWANK/BACKEND:WRAP

easye:
Bug triggered by loading McCLIM.

TODO: investigate the "right way" to initialize```swank-backend:wrap``` hooks.

### DIFF
--- a/swank/abcl.lisp
+++ b/swank/abcl.lisp
@@ -95,7 +95,10 @@
   (format stream ">")
   nil)
 
-(wrap 'sys::%print-unreadable-object :more-informative :replace '%print-unreadable-object-java-too)
+;;; TODO: move such invocations out of toplevel?  
+(eval-when (:load-toplevel)
+  (unless (get 'sys::%print-unreadable-object 'swank/backend::slime-wrap) 
+    (wrap 'sys::%print-unreadable-object :more-informative :replace '%print-unreadable-object-java-too)))
 
 (defimplementation call-with-compilation-hooks (function)
   (funcall function))


stassats:
Why does slime need that and not ABCL as a whole?

easye:
@stassats Good question.  Best answer is that we don't currently have an ADVICE-like facility in ABCL itself, but SLIME is always available via ```(asdf:load-system :swank)```.  Lemme think a bit more about this.  This patch is needed to load the McCLIM demos.

stassats:
I'm not talking about advice, but why does slime need sys::%print-unreadable-object to be '%print-unreadable-object-java-too, yet ABCL itself doesn't need that.

And as to the patch itself, you can apply it yourself, if you it's needed.

easye:
> I'm not talking about advice, but why does slime need sys::%print-unreadable-object to be '%print-unreadable-object-java-too, yet ABCL itself doesn't need that.

I'll look at what code can be profitably refactored back into the ABCL-INTROSPECT contrib to reduce this sort of thing.  

> And as to the patch itself, you can apply it yourself, if you it's needed.

I will, but I wanted to solicit feedback, which you have amply provided.  Thanks!

easye:
Merging this fix for now; filed <https://github.com/armedbear/abcl/issues/234> to track work on the elimination of this sort of code in SLIME proper.

luismbo:
@easye cool, thanks. (Minor note: we've been mostly avoiding merge commits, so next time please use the rebase and merge or squash and merge option.)

easye:
> @easye cool, thanks. (Minor note: we've been mostly avoiding merge commits, so next time please use the rebase and merge or squash and merge option.)

ACK.  

Are we moving towards a 2.25?

stassats:
Yes, 2.25 can be released at any time.

easye:
> Are we moving towards a 2.25? ABCL support has dramatically improved inspecting Java objects. I will write a CHANGES entry. And practice my hgfu with publishing properly git rebased commits.

Added as <https://github.com/slime/slime/pull/566>.

PR #567, opened 2020-06-14
slime-selector: Prefix argument means other window

lukego:
Make `slime-selector` take a prefix argument to prefer opening the buffer in the other window.

This is an equivalent alternative to the "4" selector method. Formally it is redundant but for me it rolls off the fingers a little easier, and the `slime-selector' function already had a suitable argument and only needed its `(interactive)` declaration updated.


luismbo:
Welcome back Luke.

lukego:
Thanks, and thanks everybody for keeping the project alive and thriving all these years :)

PR #570, opened 2020-06-29
slime-dispatching-connection cleanups

katterjohn:
This removes an unnecessary binding and (related) defvar.

`slime-connect` binds `slime-dispatching-connection` before calling `slime-setup-connection`, which in turn immediately binds this variable to the same value.  There were previously other functions called during the scope of this binding in `slime-connect`, but those calls have since been removed.

There has been a defvar for `slime-dispatching-connection` with an initial value and docstring that occurs after the definition of `slime-connect`.  The defvar being removed here was added later in commit 55ef9ac0 when enabling lexical binding for slime.el, and this defvar was presumably only for making the binding in `slime-connect` dynamic.  With the removal of this binding this defvar is unnecessary.

(This is just something I noticed while looking into something else.)

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -440,7 +440,6 @@ This is a hack so that we can reinitilize the real slime-mode-map
 more easily. See `slime-init-keymaps'.")
 
 (defvar slime-buffer-connection)
-(defvar slime-dispatching-connection)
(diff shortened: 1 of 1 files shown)


luismbo:
Thanks!

Issue #541, opened 2019-12-19
Travis CI tests are failing 


gendl:
Hi,

I just submitted a [PR](https://github.com/slime/slime/pull/540) and happened to notice that the Travis builds in .travis-ci.yml have been failing starting in November 2018. At least some of the failures have to do with installation of emacs on the docker image being used for the test. 

I'm not sure exactly why these are failing, but maybe someone who is more familiar could look at the .travis-ci.yml and see about getting the tests to work as a baseline? That way we can know whether new PRs are introducing regressions or not. I can try looking at it myself as well but I thought to open a bit of a discussion here first before trying to dive in (I'm still a total newbie when it comes to Travis CI and docker - the extent of my CI experience so far is just shell executor with gitlab-ci). 

Query: The latest released Gnu emacs is 26.3 - should the slime tests be including emacs 26 and maybe 25 in the tests, and maybe drop 23 at this point? 

Thanks,

 Dave

luismbo:
Yes the Travis file is in need of an upgrade.

luismbo:
The Travis build has been upgraded some time ago and now we have GitHub Actions too, although tests are still a bit finicky.

PR #576, opened 2020-07-21
[WIP] Add working CI via GitHub Actions

purcell:
Hi - I noticed your CI is failing (see #541) and I thought I'd try submitting an equivalent GitHub Actions config to see if I can help fix it. Will update here when and if I get it working. :-)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,25 @@
+name: CI
+
+on:
+  pull_request:
+  push:
+    paths-ignore:
+    - '**.md'
+
+jobs:
+  build:
+    runs-on: ${{ matrix.os }}
+    strategy:
+      matrix:
+        lisp: [ccl, sbcl]
+        os: [ubuntu-latest, macos-latest]
+        emacs_version: [24-5, 25-3, 26-3, snapshot]
+    steps:
+
+    - uses: cachix/install-nix-action@v8
+    - uses: purcell/setup-emacs@master
+      with:
+        version: ${{ matrix.emacs_version }}
+    - uses: actions/checkout@v2
+    - run: nix-env -i ${{ matrix.lisp }}
+    - run: make LISP="${{ matrix.lisp }}" check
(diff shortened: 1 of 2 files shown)


purcell:
That's mostly working:

* CCL and SBCL
* Linux and MacOS
* Same selection of Emacs versions as before, plus the latest HEAD

The Linux builders run in < 2mins, the MacOS ones fail, and I suspect there may be a timeout which needs extending? We could remove that platform if desired, but it's kinda nice to have cross-platform checks.

(I've added a build status badge to the README, and if you eventually merge this, you might also consider removing the travis badge and config to discontinue its use.)

stassats:
I disabled the macos target, but the some tests are still flaky.

luismbo:
The Travis build is actually passing, the badge is just not being updated for some reason. In any case, GitHub Actions sounds good. I'll remove the Travis setup. Thanks @purcell.

luismbo:
@stassats the nix packages for SBCL seem a bit out-of-date, do we care?

stassats:
It's not that old, the one shipped with ubuntu is much older, so, short of building it manually, this should be ok.

luismbo:
OK. If we ever run into trouble with the nix packages we can always revert back to the roswell way: `curl -L https://raw.githubusercontent.com/snmsts/roswell/release/scripts/install-for-ci.sh | sh` and `LISP=ros run -L $LISP`. Removed Travis CI in 5f48c8c1f3c1538cccc155268d304172e6339aa7.

purcell:
Thanks for the merge, folks!

> OK. If we ever run into trouble with the nix packages we can always revert back to the roswell way

Yeah, I figured try the nix way, since that's what the `setup-emacs` action uses under the hood. It should track the Nix unstable channel by default, so package versions should stay pretty up-to-date overall.

luismbo:
@purcell the nix step has stopped working. Any idea why?

purcell:
Fixed in #593

luismbo:
Thanks @purcell!

purcell:
No worries, thanks for tagging me in.

PR #582, opened 2020-10-20
Image buffers

christophejunke:
First patch allows custom message to pass through swank.lisp without error.
The second one allows to pass base64 strings as :data directly

```
(swank::send-to-emacs
          `(:write-image (:data ,(base64:usb8-array-to-base64-string
                                  (vecto:with-canvas (:width 64 :height 64)
                                    (vecto:set-rgb-stroke 100 100 100)
                                    (vecto:move-to 0 0)
                                    (vecto:line-to 64 64)
                                    (vecto:stroke)
                                    (flexi-streams:with-output-to-sequence (out)
                                      (vecto:save-png-stream out)))))
            " "))
```

### DIFF
--- a/contrib/slime-media.el
+++ b/contrib/slime-media.el
@@ -11,25 +11,35 @@
    (add-hook 'slime-event-hooks 'slime-dispatch-media-event)))
 
 (defun slime-media-decode-image (image)
-  (mapcar (lambda (image)
-	    (if (plist-get image :data)
-		(plist-put image :data (base64-decode-string (plist-get image :data)))
-	      image))
-	  image))
+  (typecase image
+    (string (list :file image :type nil))
+    (list
+     (if (plist-get image :data)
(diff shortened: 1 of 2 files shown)


luismbo:
Is the intent here to be able to write to the repl buffer only or to other buffers as well?

I have two improvement suggestions. Add a higher-level function to slime-media swank contrib to avoid sending swank messages directly. Use ed-rpc rather than removing the swank side message validation (or add a specific handler, but I think the ed-rpc way is nicer).

christophejunke:
> Is the intent here to be able to write to the repl buffer only or to other buffers as well?

other buffers as well, if possible

> Add a higher-level function to slime-media swank contrib to avoid sending swank messages directly. 

Something like `swank:send-image`?

> Use ed-rpc rather than removing the swank side message validation (or add a specific handler, but I think the ed-rpc way is nicer).

I don't really know what ed-rpc is so I'll have a look.

Thank you

PR #611, opened 2021-04-15
Prevent creating a letftover buffer when loading the symbol table

Uthar:
When `common-lisp-hyperspec--parse-map-file` is run, it leaves open a buffer with the Map_Sym.txt file from the Hyperspec. By using `with-temp-buffer`, this patch prevents it.

### DIFF
--- a/lib/hyperspec.el
+++ b/lib/hyperspec.el
@@ -138,7 +138,7 @@ If you copy the HyperSpec to another location, customize the variable
 
 (defun common-lisp-hyperspec--parse-map-file (file)
   (with-temp-buffer
-    (insert-file-contents-literally file)
+    (insert-file-contents file)
     (goto-char (point-min))
     (let ((result '()))
       (while (< (point) (point-max))
--- a/lib/hyperspec.el
+++ b/lib/hyperspec.el
@@ -137,7 +137,8 @@ If you copy the HyperSpec to another location, customize the variable
     (forward-line)))
 
 (defun common-lisp-hyperspec--parse-map-file (file)
-  (with-current-buffer (find-file-noselect file)
+  (with-temp-buffer
+    (insert-file-contents-literally file)
     (goto-char (point-min))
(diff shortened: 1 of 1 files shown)


lib/hyperspec.el
```diff
@@ -137,7 +137,8 @@ If you copy the HyperSpec to another location, customize the variable
     (forward-line)))
 
 (defun common-lisp-hyperspec--parse-map-file (file)
-  (with-current-buffer (find-file-noselect file)
+  (with-temp-buffer
+    (insert-file-contents-literally file)
```
luismbo, line 141: Maybe `insert-file-contents` would be safer here, should the symbol table file not be in the same encoding as the default buffer encoding? E.g. if the symbol table is in CRLF format, symbol names will be parsed with a trailing CR.

luismbo (changes requested):
I'm curious about the motivation for this. Do you parse symbol tables on a regular basis rather than use the builtin-in index?

lib/hyperspec.el
Uthar: Yeah, that's a bug - fixed in 44126138b24a08b665aacc981be09fd37dd8746f

Uthar:
> I'm curious about the motivation for this. Do you parse symbol tables on a regular basis rather than use the builtin-in index?

I just use those `parse` functions to make `slime-hyperspec-lookup` work with HyperSpec-4-0, because the hard-coded values assume a newer HyperSpec. 
To do that I just set these 2 variables:
```
(setq common-lisp-hyperspec-root "@clhs@/")
(setq common-lisp-hyperspec-symbol-table "@clhs@/Data/Symbol-Table.text")
```

luismbo:
> I just use those `parse` functions to make `slime-hyperspec-lookup` work with HyperSpec-4-0, because the hard-coded values assume a newer HyperSpec.

👍 OK, thanks!

PR #618, opened 2021-05-03
Fix up removal of cl package dependency in slime-editing-commands.el

rhgg2:
Some `incf`s and `decf`s should have been `cl-incf`s and `cl-decf`s in `slime-editing-commands.el`.


luismbo:
Thanks @rhgg2!

PR #625, opened 2021-07-16
Add keyword-args to macro "define-minor-mode" to eliminate deprecation

wlharvey4:
The macro "define-minor-mode" takes keyword arguments rather than
positional arguments, and a "deprecation" warning is given wherever the keyword
arguments are missing.  These changes add keyword arguments where they are missing
to eliminate the warnings.


luismbo:
Thanks @wlharvey4!

Issue #286, opened 2015-12-18
SLIME is insecure and easily allows local attackers to execute arbitrary code


DemiMarie:
SLIME is extremely insecure and allows local attackers to easily execute arbitrary code by connecting to the Swank server process.

Fortunately, the fix is simple, at least on Unix: use a Unix domain socket (`AF_UNIX`) instead of a TCP socket on `localhost`.  On Windows, use a simple, once-only secret key obtained from `RtlGenRandom()` (the basis of the better-known, and harder to use, `CryptGenRandom()`) to authenticate Emacs to the Common Lisp program.

The secret should be passed via a secure temporary file.  Specifically, Emacs should set the file contents to the authentication key, and the Common Lisp implementation should overwrite the file with its port number.  One MUST use a means of comparison that resists timing attacks (attackers can hammer `localhost` very quickly).

One could also use the Windows solution on Unix, with `/dev/urandom` as the random number source.

luismbo:
I suppose the vast majority of SLIME users are on single-user machines. Definitely something we should do at some point, though. https://github.com/slime/slime/pull/270 is a first stab at implementing your AF_UNIX suggestion, but it might be a better idea to use the same solution on Windows and UNIX for better maintainability.

drbo:
I have come up with a plan of attack.

Part 1 is to enable use of Unix domain sockets.  Much of this code can be taken from the Emacs server code within Emacs itself.  This fixes the problems forever on must Unix systems, unless ABCL or LispWorks is used.  ABCL, being a JVM language, does not support Unix domain sockets without a JNI library.  I couldn't figure out how to change the LispWorks code since it uses undocumented functions and I don't have LispWorks.

Part 2 is securing transport over TCP sockets.  My plan here is, essentially, to make the existing `.slime-secret` support much more secure.  Instead of one `.slime-secret` file, there will be one `.slime/secret_keys/pid` file for each instance of Swank.  Swank and Emacs will both check permissions on these files.  Two passwords will be generated by Swank at server startup using a cryptographically secure random number generator, and will each be 32 bytes long.  One will authenticate Emacs to Swank, and the other will authenticate Swank to Emacs.  Neither password will ever be reused.  Also, Swank and Emacs will both be hard-coded to listen to `localhost` &mdash; Swank has no business listening on a non-`localhost` network connection.  Remote connections can be taken care of via SSH and socat.

zellerin:
[...]

PR #635, opened 2021-08-31
UNIX domain socket support in a non-exclusive way

ssafar:
Opening TCP sockets, even locally, is somewhat tricky from a security point of view (see #286)... especially when connecting to remote Lisps when we don't want to close the port immediately.

#291 was aiming to solve this problem by switching over to UNIX domain sockets entirely when connecting to Lisps on UNIX systems, and sticking with TCP ports e.g. on Windows. However, this doesn't account for cases when we'd still want to keep using TCP ports on UNIX; also, its author is not interested in working on this anymore.

This PR follows a slightly different approach: instead of replacing the existing, TCP-based connection mechanism, it adds UNIX domain sockets as an additional, optional mechanism. We still connect to local Lisps via TCP; the use case for this is long-running Lisp images on servers, that we'd want to connect to remotely (e.g. by using ssh to forward the UNIX socket, or wrapping it into a client-cert-checked SSL tunnel using e.g. socat). It's definitely possible to use this sort of mechanism for local Lisps too in the future; it's just not being done yet, to keep things simple (& avoid breaking too many things).

I also added a test to make sure that this does in fact work, end to end.

It currently only supports SBCL; I can definitely put in backend functions for other Lisps, too, before merging, if you think this entire PR is a good idea in general (... if it isn't, universal support is kinda pointless :) so the first commit is mainly to start a discussion.) 

Thanks for taking a look!

### DIFF
--- a/packages.lisp
+++ b/packages.lisp
@@ -137,6 +137,7 @@
   (:export #:startup-multiprocessing
            #:start-server
            #:create-server
+           #:create-server-unix
            #:stop-server
            #:restart-server
            #:ed-in-emacs
--- a/slime-tests.el
+++ b/slime-tests.el
@@ -1292,10 +1292,29 @@ This test will fail more likely before dispatch caches are warmed up."
                                              :dont-close nil)))))
            (slime-sync-to-top-level 3)
            (slime-disconnect)
-           (slime-test-expect "Number of connections must remane the same"
+           (slime-test-expect "Number of connections must remain the same"
                               connection-count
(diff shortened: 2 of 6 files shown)


zyd-n:
@ssafar 

> It's definitely possible to use this sort of mechanism for local Lisps too in the future; it's just not being done yet, to keep things simple (& avoid breaking too many things).

Any pointers on what's needed for adding support for local setups? I'm interested.

PR #640, opened 2021-11-07
See if more can get into slime

alanruttenberg:
I rebased whatever hadn't been brought into slime to head. I don't expect you to pull as is, but should be easy to see what I've done and discuss what might be acceptable.


alanruttenberg:
There are test failures, but all it says is "Error:" How does one debug this?

alanruttenberg:
The intent was to have this be a pull to Evenson's version. I'm closing this one pending working things out there.

PR #648, opened 2021-12-15
Simplify message timestamp formatting

eggert:
* slime-tests.el (slime-wait-condition): Use format-time-string rather
than formatting by hand.  This avoids relying on Emacs timestamp
format, which is slated to change.

### DIFF
--- a/slime-tests.el
+++ b/slime-tests.el
@@ -200,9 +200,8 @@ conditions (assertions)."
 (defun slime-wait-condition (name predicate timeout)
   (let ((end (time-add (current-time) (seconds-to-time timeout))))
     (while (not (funcall predicate))
-      (let ((now (current-time)))
-        (message "waiting for condition: %s [%s.%06d]" name
-                 (format-time-string "%H:%M:%S" now) (cl-third now)))
+      (message "waiting for condition: %s [%s]" name
+               (format-time-string "%H:%M:%S.%6N"))
       (cond ((time-less-p end (current-time))
              (error "Timeout waiting for condition: %S" name))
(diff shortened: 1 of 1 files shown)


monnier:
Hmm.. important note: this is not the right commit to merge, because it has commit ID 146837d, whereas the commit that was mistakenly pushed to `nongnu.git` is commit 1923227f54.  Please do:

    git fetch git://git.sv.gnu.org/emacs/nongnu.git elpa/slime

then you can check

   git show FETCH_HEAD

or

    git show 1923227f54

monnier:
Ping?
This problem is still preventing new releases of SLIME on NonGNU ELPA.

luismbo:
@monnier I don't understand how merging that commit fixes anything. The resulting history in slime.git would still be different from nongnu.git's, wouldn't it?

monnier:
That `nongnu.git` is different is not a problem: it's the normal case whenever someone commits a change to Slime, after all.  After that `nongnu.git` catches up with a "fast forward".

With this merge, `slime.git` would still be different, yes, but it would go back to being a normal kind of different, i.e. "strictly ahead" (because the head of `nongnu.git` would be included as an ancestor of the head of `slime.git`), which is the normal case which can be resolved trivially with a fast-forward.

monnier:
Ping?  NonGNU ELPA is trailing something like 30 commits behind slime.git now :-(

monnier:
Ping? Ping? NonGNU ELPA is now a year behind!

stassats:
Take it up with NonGNU then.

PR #650, opened 2021-12-22
Use xref-push-marker-stack when available

theothornhill:
Inserting directly into `find-tag-marker-ring` is obsolete since emacs 25.1.  As
of emacs 29 slime no longer is able to jump back and forth between definition
and usage with `M-.` and `M-,`.  As a precaution for backwards compatibility, we
use `xref-push-marker-stack` only when it is bound.


stassats:
Looks like it was fixed in 789584a7acb15747678fa62a8fcfc8d1187be867

PR #700, opened 2022-01-12
Fix find-definition stack not working in upcoming emacs versions

### LINKED ISSUE #699: `slime-pop-find-definition-stack' not working on recent emacs master
Because `find-tag-marker-ring' is obsolete since 25.1 inserting the point marker on M-. does nothing anymore.

astrangeguy:
This fixes #699

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -3672,14 +3672,14 @@ alist but ignores CDRs."
 ;;;; Edit definition
 
 (defun slime-push-definition-stack ()
-  "Add point to find-tag-marker-ring."
-  (require 'etags)
-  (ring-insert find-tag-marker-ring (point-marker)))
+  "Add point to find-tag-marker-stack."
+  (require 'xref)
+  (xref-push-marker-stack (point-marker)))
 
 (defun slime-pop-find-definition-stack ()
   "Pop the edit-definition stack and goto the location."
   (interactive)
-  (pop-tag-mark))
+  (xref-pop-marker-stack))
 
 (cl-defstruct (slime-xref (:conc-name slime-xref.) (:type list))
   dspec location)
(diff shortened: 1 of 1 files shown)


slime.el
```diff
@@ -3672,14 +3672,14 @@ alist but ignores CDRs."
...
-  (ring-insert find-tag-marker-ring (point-marker)))
+  "Add point to find-tag-marker-stack."
+  (require 'xref)
+  (xref-push-marker-stack (point-marker)))
 
 (defun slime-pop-find-definition-stack ()
   "Pop the edit-definition stack and goto the location."
   (interactive)
-  (pop-tag-mark))
+  (xref-go-back))
```
luismbo, line 3682: `xref-go-back` doesn't exist in, e.g., Emacs 27.1

luismbo: I guess we should use `xref-pop-marker-stack` instead?

astrangeguy: Yes, stupid of me not to notice the old name 3 lines above...

slime.el
```diff
@@ -3672,14 +3672,14 @@ alist but ignores CDRs."
 ;;;; Edit definition
 
 (defun slime-push-definition-stack ()
-  "Add point to find-tag-marker-ring."
-  (require 'etags)
-  (ring-insert find-tag-marker-ring (point-marker)))
+  "Add point to find-tag-marker-stack."
+  (require 'xref)
```
luismbo, line 3676: While we’re at it. Why require at runtime? Can’t we just move this dependency to the beginning of the file like every other dependency?

astrangeguy: Seems to be an artifact of something from the past... the runtime-require of etags is unnecessary since it's a dependency ether way

astrangeguy: Yeah it's from when etags was loaded only when needed on XEmacs

luismbo:
@astrangeguy great, thanks. 👍

PR #706, opened 2022-02-03
Don't allow duplicate entries in the source file cache.

melisgl:
It's an EQUAL hash table, so make sure NAMESTRINGs are converted to
PATHNAMEs. This assumes that the filename is never NIL.


stassats:
I'm not sure why it needs to be converted to pathnames?

melisgl:
`(equal #P"xxx" "xxx")` is false,  so a caller passing a namestring leads to a cache miss, a subsequent read of the file, and finally `*source-file-cache*` having both `#P"xxx"` and `"xxx"` as keys with the same potentially large string (of the file contents).
I considered fixing the call sites, but that would involve some tedium, and the same issue may be reintroduced later.

stassats:
Ok, so the types are mixed, now it makes more sense.

PR #708, opened 2022-02-10
swank.asd: Do not reload SWANK.

melisgl:
Previously, it was the case that starting SLIME in Emacs loaded Swank
with swank-loader.lisp. If Swank was then loaded via asdf (e.g. by
loading a system that depends on it), then swank-loader.lisp was run
again leading to lots of redefinition warnings and broke SLIME if the
version of Swank loaded by asdf was not compatible with the already
running version.

This change makes swank.asd never reload Swank. In particular, if it
finds that Swank is already loaded (because the SWANK-LOADER package
exists), then it compares the TRUENAMEs of the directories of the two
versions of Swank. If they are the same, then it silently does
nothing. If they are not the same, then it still does not reload
Swank, but it now emits a warning.

There is an unimportant corner case. Suppose the two versions are the
same, the sources of Swank change, and a reload is required. With this
change, swank-loader.lisp is to be invoked explicitly. Previously,
loading Swank via asdf _for the first time_ reloaded Swank and picked
up the changes, and subsequent attempts to load it via asdf silently
did nothing (unless swank-loader.lisp itself was changed).

### DIFF
--- a/swank.asd
+++ b/swank.asd
@@ -6,6 +6,13 @@
 ;; This is only useful if you want to start a Swank server in a Lisp
 ;; processes that doesn't run under Emacs. Lisp processes created by
 ;; `M-x slime' automatically start the server.
+;;
+;; If Swank is already loaded (e.g. the Lisp is running under SLIME),
+;; then attempts to load it via asdf do nothing, except for emitting a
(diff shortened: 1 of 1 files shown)


luismbo:
👍

PR #713, opened 2022-02-27
Handle null characters correctly in the Allegro backend.

brown:
The Allegro function EXCL:OCTETS-TO-STRING, at least in Free Express Edition
10.1, stops processing octets when it encounters a null character, unless the
target of octet conversion is an existing string.  Work around the behavior
in UTF8-TO-STRING.

Modify test utf-8-source in slime-tests.el so that it includes a null
character.


luismbo:
Merged, thanks!

Issue #437, opened 2018-03-20
prettify-symbols-mode breaks SLIME REPL coloring


TeMPOraL:
It seems that activating prettify-symbols-mode (Emacs builtin) breaks fontification in SLIME. It also has the consequence of completely breaking the in-REPL coloring and clickable areas introduced with log4slime from [log4cl project](https://github.com/sharplispers/log4cl).

Steps to reproduce:

* Open Emacs (e.g. with -Q for clear config).
* Ensure SLIME is loaded with `(slime-setup '(slime-fancy))`
* Open SLIME with `M-x slime`
* Activate prettify-symbols-mode with `M-x prettify-symbols-mode`
* Colors in SLIME REPL should disappear, and will no longer appear until prettify-symbols is disabled and SLIME is `,quit` and reopened.

Version info:

* GNU Emacs 25.2.2 (x86_64-pc-linux-gnu, GTK+ Version 3.22.21) of 2017-09-22, modified by Debian
* slime-20171207.1712
* Ubuntu 17.10

ghollisjr:
I can confirm this bug is still active.  Some clues I've found:

* SLIME uses add-text-properties and font-lock-keyword-face to colorize the REPL (slime-repl.el: 589)
* prettify-symbols-mode uses font-lock-remove-keywords to clear highlighting (prog-mode.el: 217)
* Before running `prettify-symbols-mode`: `(text-properties-at 15 (get-buffer "*slime-repl sbcl*"))` ==> `(fontified t field output inhibit-line-move-field-capture t front-sticky (read-only) rear-nonsticky t slime-repl-prompt t read-only t face slime-repl-prompt-face)`
* After running `prettify-symbols-mode`: `(text-properties-at 15 (get-buffer "*slime-repl sbcl*"))` ==> `(fontified t field output inhibit-line-move-field-capture t front-sticky (read-only) rear-nonsticky t slime-repl-prompt t read-only t)`

So it clearly looks like `prettify-symbols-mode` is failing to set the face property for some reason.  I don't yet know enough to say whether this is something SLIME needs to address or if it's an upstream bug in `prettify-symbols-mode`.

ghollisjr:
More clues:

* `slime-repl-insert-prompt` is still trying to set the `face` text property when `prettify-symbols-mode` is enabled, but it apparently fails somehow as that property is never added to the text properties for the prompt.
[...]

PR #717, opened 2022-04-13
Fixed prettify-symbols-mode bug

ghollisjr:
Fix for the bug described in [Issue #437](https://github.com/slime/slime/issues/437)


luismbo:
Could you explain, perhaps in the commit message, what is going on? TIA.

ghollisjr:
Unfortunately not, the issue is that `prettify-symbols-mode` uses `font-lock-mode` to manage symbol substitution & highlighting, and SLIME functions don't use the `font-lock-mode` hook-setting variables, which are usually 1-to-1 replacements.

So I added those variables to the SLIME functions and now SLIME works with or without `prettify-symbols-mode`/`font-lock-mode`.

I mentioned a small blemish in the REPL coloring when using `prettify-symbols-mode`, but that's specifically a `prettify-symbols-mode` issue and can't be helped from SLIME's side.

PR #723, opened 2022-05-10
Properly comment out multiline error messages during printout

### LINKED ISSUE #721: Improper partial commenting of sbcl errors causes indentation to fail.
Hello world,

I am currently troubleshooting a bug in slime that is breaking indentation in the repl.

As an example, if I open an s-expr in the repl,

```
CL-USER> (my-function
   second-line
```
and I press tab anywhere on the second line, the text is automatically indented like so

```
CL-USER> (my-function
          second-line
```

However, when a multiline error message is printed, for example by a simple call to `(assert (= 1 0))`, this no longer happens.
```
; SLIME 2.27
CL-USER> (assert (= 0 1))
; Evaluation aborted on #<SIMPLE-ERROR "~@<The assertion ~S failed~:[.~:; ~
                                           with ~:*~{~S = ~S~^, ~}.~]~:@>" {1004702153}>.
CL-USER> (my-function
   second-line
```
Putting the cursor on the second-line and pressing tab will now bring up the completion menu instead of indenting the line.

I think that this is caused by the second line. I think the indentation-maker is evaluating the code like the commented out line doesn't exist, and so the `with ~:*~{~S = ~S~^, ~}.~]~:@>" {1004702153}>.` is opening a string, which needs to be closed before pressing tab will indent further lines.

For example, modifying the text already printed in the buffer with the following will restore indentation:
```
CL-USER> (assert (= 1 2))
; Evaluation aborted on #<SIMPLE-ERROR "~@<The assertion ~S failed~:[.~:; ~
;                                           with ~:*~{~S = ~S~^, ~}.~]~:@>" {1004A89B63}>.
CL-USER>
```
alternatively, adding an end-quote will also restore the ability to indent code. (i.e. `~}.~]~:@>"` becomes `~}.~]~:@>""`.)

```
CL-USER> (assert (= 1 2))
; Evaluation aborted on #<SIMPLE-ERROR "~@<The assertion ~S failed~:[.~:; ~
                                           with ~:*~{~S = ~S~^, ~}.~]~:@>"" {1004A89B63}>.
CL-USER>
```

Given that this bug is heavily annoying to my coworkers, I'm willing to fix it, but am completely lost as to where this `;` is getting added.

Patrick-Poitras:
This closes #721, in which multi-line error messages could get written to the repl without the second line being commented. This made the indentation fail in some cases.

### DIFF
--- a/contrib/slime-repl.el
+++ b/contrib/slime-repl.el
@@ -578,8 +578,10 @@ joined together."))
       (slime-save-marker slime-output-start
         (slime-save-marker slime-output-end
           (goto-char slime-output-end)
-          (insert-before-markers (format "; Evaluation aborted on %s.\n"
-                                         condition))
+          (insert-before-markers
+           ;; Comment-out multi-line error messages.
+           (format "; Evaluation aborted on %s.\n"
+                   (replace-regexp-in-string "\n" "\n; " condition)))
           (slime-repl-insert-prompt))))
     (slime-repl-show-maximum-output)))
 


stassats:
Can it be
```
(format "; Evaluation aborted on %s.\n"
        (replace-regexp-in-string "\n" "\n; " condition))
```
instead?

Patrick-Poitras:
@stassats I have made the change you suggested. Thank you for the help!

PR #724, opened 2022-05-14
sbcl: Improve inspection of functions and code components.

karlosz:
* Restore display of entry points for code components.
* Show disassembly for simple funs.


luismbo:
Thanks!

Issue #583, opened 2020-10-21
Not using debug-source-created in SBCL backend?


glv2:
Would it be possible to remove the use of ```sb-c::debug-source-created``` in the SBCL backend and to get the date from the filesystem instead?

This could help improve the reproducibility of fasl files compiled by SBCL (see https://sourceforge.net/p/sbcl/mailman/message/37133598/ and https://sourceforge.net/p/sbcl/mailman/message/37133640/).

luismbo:
Yes, that seems reasonable.

PR #726, opened 2022-05-22
sbcl: Use file-write-date instead of debug-source-created.

glv2:
Would this patch or a similar one be enough to fix issue #583 ?

### DIFF
--- a/swank/sbcl.lisp
+++ b/swank/sbcl.lisp
@@ -1349,8 +1349,8 @@ stack."
       (fallback-source-location code-location)))
 
 (defun source-file-source-location (code-location)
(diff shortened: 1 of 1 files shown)


luismbo:
Thanks.

luismbo:
This however makes `swank/source-file-cache` useless on SBCL so it'd be nice to clean that up and not use it on SBCL.

luismbo:
Or maybe it'd be nicer to only fall back to `file-write-date` when `debug-source-created` has been removed. I may have merged this prematurely. :-/

luismbo:
I've reverted 3837255 for now. Sorry for the noise. I think this needs to be figured out in SBCL before we adapt SLIME.

PR #729, opened 2022-07-09
clasp: Replace TMP host with EXT:TEMPORARY-DIRECTORY when available

yitzchak:
Clasp doesn't have the TMP logical host anymore, instead there is a function `ext:temporary-directory`,


drmeister:
I approve this change.

PR #735, opened 2022-09-02
slime-presentations: handle *read-base* not being 10

deftransform:


stassats:
There's already a dot at the end, so why is `#10r` needed?

deftransform:
You are probably right.

I don't remember now why I did that but I will try to see if I can find a counterexample.

PR #739, opened 2022-11-01
Add setting to enable auto-ignore errors on slime-repl-history io failures

Patrick-Poitras:
We have been getting a few problems where a failure of the REPL loading/saving would prompt the user for input and some automated workflows require that this error, if it occurs, be allowed to be ignored without prompting.

This commit adds a new setting `slime-repl-history-ignore-io-errors` at the level of the `slime-repl-call-with-handler` function. This function is only used in file IO for the REPL history, so I've renamed this function `slime-repl-history-call-with-handler` as to make it obvious that it isn't used elsewhere. I've also cleaned up the function a little bit.

Cheers!


stassats:
Silently ignoring errors doesn't sound great.

Patrick-Poitras:
@stassats Yeah, I think that most users would like to have an error message at least tell them if the situation happened. It's not something our (i.e. internal to my workplace) users care about, but as a general interface, it's not great that it would immediately default to ignoring error messages.

I've made a second version that warns the user if it happens. I'm much happier with it from an UX perspective, thank you so much for the suggestion.

PS. Sorry for the multiple overwriting builds in the last hour, there was an error that triggered on CCL but not on SBCL and I was trying to see if it had to do with the function return being different. Forgot this is connected and updates automatically. My apologies if you got pinged 50 times.

PR #741, opened 2022-11-19
Fix slime-load-system for systems with dots in the name

### LINKED ISSUE #609: Call to PATHNAME-NAME in SWANK:LIST-ALL-SYSTEMS-IN-CENTRAL-REGISTRY gives wrong result for system names containing '.'
Hello,

I have an ASDF system called "com.silentflame.consfig".  The function SWANK:LIST-ALL-SYSTEMS-IN-CENTRAL-REGISTRY maps PATHNAME-NAME over all system names and thus corrupts "com.silentflame.consfig" to "com.silentflame" which is a system which does not exist.  This makes it difficult to use the repl shortcut for ASDF:LOAD-SYSTEM to load my system, because "com.silentflame" shows up as a completion candidate but that would not be a valid system to load.

Thanks.

Uthar:
This was causing wrong results for system names containing '.' when using the load-system shortcut

Test like this:

1. `mkdir -p ~/common-lisp/foo.bar.baz/src/`
2. `touch ~/common-lisp/foo.bar.baz/src/foo.bar.baz.asd`
3. `M-x slime`
4. `M-x slime-load-system foo TAB`
5. `foo.bar.baz` should appear in the `*Completions*` buffer

Fixes #609

### DIFF
--- a/contrib/swank-asdf.lisp
+++ b/contrib/swank-asdf.lisp
@@ -386,24 +386,25 @@ Example:
   "Returns a list of all systems in ASDF's central registry
 AND in its source-registry. (legacy name)"
   (unique-string-list
-   (mapcar
-    #'pathname-name
-    (while-collecting (c)
-      (loop for dir in asdf:*central-registry*
-            for defaults = (eval dir)
-            when defaults
-            do (collect-asds-in-directory defaults #'c))
-      (asdf:ensure-source-registry)
-      (if (or #+asdf3 t
-	      #-asdf3 (asdf:version-satisfies (asdf:asdf-version) "2.15"))
-          (loop :for k :being :the :hash-keys :of asdf::*source-registry*
-		:do (c k))
-	  #-asdf3
-          (dolist (entry (asdf::flatten-source-registry))
-            (destructuring-bind (directory &key recurse exclude) entry
-              (register-asd-directory
-               directory
-               :recurse recurse :exclude exclude :collect #'c))))))))
+   (while-collecting (c)
+     (loop for dir in asdf:*central-registry*
+           for defaults = (eval dir)
+           when defaults
+           do (collect-asds-in-directory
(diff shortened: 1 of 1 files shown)


contrib/swank-asdf.lisp
```diff
@@ -387,7 +387,11 @@ Example:
 AND in its source-registry. (legacy name)"
   (unique-string-list
    (mapcar
-    #'pathname-name
+    (lambda (asd)
+      (typecase asd
+        (pathname (pathname-name asd))
+        (string asd)
+        (t asd)))
```
luismbo, line 394: instead of this typecase, could we perhaps consistently produce strings from the collection below?

Uthar: How about this, producing pathnames? https://github.com/slime/slime/commit/2fdbad58ac801f98d4f7e4aa3bf1304e26633ce1

luismbo: How about using `:collect (lambda (path) (c (pathname-name path)))`?  Does that work?

Uthar: It would also work
But it looks simpler with collecting strings, take a look at the last commit

contrib/swank-asdf.lisp
```diff
@@ -386,13 +386,14 @@ Example:
   "Returns a list of all systems in ASDF's central registry
 AND in its source-registry. (legacy name)"
   (unique-string-list
-   (mapcar
-    #'pathname-name
-    (while-collecting (c)
+   (while-collecting (c)
```
luismbo, line 389: This form moved left one column, so its body needs to be reindented. Other than that this looks great.

Uthar: indented now

luismbo:
Merged, thanks!

Issue #744, opened 2022-12-02
Incorrect indentation in lambda lists for define-compiler-macro and define-setf-expander


galdor:
With `lisp-indent-function` being `common-lisp-indent-function`, provided by Slime in `slime-cl-indent.el`, the following code is indented in a strange way:

```lisp
(define-compiler-macro foo (&whole form
                                   a b c
                                   &optional (d 0))
  form)
```

One would expect the following:

```lisp
(define-compiler-macro foo (&whole form
                            a b c
                            &optional (d 0))
  form)
```

The problem is the same with `&environment` in `define-setf-expander`.

`&whole` and `&environment` can only be followed by a single value; additional values are unrelated to them and should not be indented as so.

Looking at `common-lisp-init-standard-indentation`, it seems that there are no specific rules for `define-compiler-macro`, and that `define-setf-expander` is indented as `defun`.

I'm not sure how to fix this. Does anyone has an idea?

Context:
- Emacs 28.2.
- Slime 2.27.

luismbo:
This seems to achieve your expected behaviour.

```diff
diff --git a/contrib/slime-cl-indent.el b/contrib/slime-cl-indent.el
index a620b691..7a6e4093 100644
--- a/contrib/slime-cl-indent.el
+++ b/contrib/slime-cl-indent.el
@@ -1054,7 +1054,7 @@ defun lisp-indent-report-bad-format
 (defvar lisp-indent-lambda-list-keywords-regexp
   "&\\(\
-optional\\|rest\\|key\\|allow-other-keys\\|aux\\|whole\\|body\\|\
+optional\\|rest\\|key\\|allow-other-keys\\|aux\\|body\\|\
 environment\\|more\
 \\)\\>"
   "Regular expression matching lambda-list keywords.")
@@ -1730,6 +1730,7 @@ defun common-lisp-init-standard-indentation
              (defcustom   (4 2 2 2))
              (defparameter     (as defvar))
              (defconst         (as defcustom))
+             (define-compiler-macro (as defun))
              (define-condition (as defclass))
              (define-modify-macro (4 &lambda &body))
              (defsetf      lisp-indent-defsetf)
```

galdor:
Thank you!

I submitted a pull request to fix both issues. Regarding &WHOLE and &ENVIRONMENT, I did fix the indentation algorithm instead of just removing them from `lisp-indent-lambda-list-keywords-regexp` because the later felt like a hack which could have unintended consequences.

I'll close the issue once the PR is merged.

PR #745, opened 2022-12-04
Improve lambda list indentation

galdor:
These patches should fix the issue reported in #744.

More comments in commit messages.


contrib/slime-cl-indent.el
```diff
@@ -1729,6 +1729,7 @@ Cause subsequent clauses to be indented.")
              (defconstant (as defvar))
              (defcustom   (4 2 2 2))
              (defparameter     (as defvar))
+             (define-compiler-macro (as defun))
```
luismbo, line 1744: I think these are supposed to be sorted alphabetically. If you find others that are out of order and want to sort them please make that a different commit.

galdor: Done

contrib/slime-cl-indent.el
```diff
@@ -1116,14 +1120,22 @@ environment\\|more\
...
-                        next nil)
-                (setq indent col))))
+              (cond
+               ((looking-at lisp-indent-lambda-list-single-arg-keywords-regexp)
+                ;; Some keywords such as &whole have a single argument;
+                ;; following arguments are indented to the beginning of the
+                ;; lambda-list.
+                (setq indent col
+                      next nil))
+               ((looking-at lisp-indent-lambda-list-keywords-regexp)
```
luismbo, line 1130: There's another reference to `lisp-indent-lambda-list-keywords-regexp` earlier in this function. Does it need to be adapted too?

galdor: I do not think so; I tried lots of adjustments because the function is really opaque, and this is how I got indentation which works as intended in without affecting other lambda lists.

I'm pretty sure that there are other edge cases (e.g. the fact that &ALLOW-OTHER-KEYS does not take any argument), but at this point I am just trying to fix a very specific problem without breaking anything.

luismbo: ACK. Sounds reasonable.

luismbo:
Thanks.

PR #748, opened 2023-01-12
add a suffix to the name of inferior lisp buffers

galdor:
In the current state, SLIME names all inferior lisp buffers `*inferior-lisp*` by default. When starting multiple implementations, we end up with multiple inferior lisp buffers whose only difference is a non-descriptive number suffix.

With this patch, we use the name of the implementation as suffix, in a way similar to REPL buffers.

For example with two inferior lisps SBCL and CCL, we now have `*inferior-lisp sbcl*` and `*inferior-lisp ccl*`


galdor:
Could someone have a look and merge the patch if it is ok? It really helps when you are have multiple connections open (e.g. when checking code against multiple implementations).

luismbo:
There's a concept of connection name (the `:name` argument to `slime-start` IIRC). Perhaps it'd be better to use that.

galdor:
You mean use what was passed with `:name` instead of using the name of the implementation? I can't seem to find how `:name` is usually being passed when running `slime`, is it even set somewhere?

luismbo:
Yes, I think it's only possible to set it via `slime-start`. Perhaps the `slime` command should provide a better default: the implementation name, as you suggest. (Another option would be to use the core/image name, perhaps without file type.) The name needs to be passed down to `slime-start-lisp` too, since there's no `slime-connection-name` at that point yet.

galdor:
Yes it could make sense. I'll rework this patch when I get the chance.

PR #760, opened 2023-03-08
Rework swank.asd to produce actual compilation artifact

dkochmanski:
Until now swank.asd compiled the system files when it was loaded. That is roughly compatible with what slime does when loading swank, however that is not useful for swank that is loaded as a standalone system without slime. Instead of producing compilation artifacts that may be further processed by asdf (either in bundle-op or otherwise) it puts fasls in ~/.slime directory.

When a system "foo" depends on "swank" then before the system "foo" is loaded swank is loaded itself. However if we only compile systems and then try to load a bundle then all we have is the swank-loader artifact - no "SWANK" package nor nothing. This issue has been unnoticed because usually people dump images instead of having separate compilation and loading steps.

This commit defines two systems "swank" and "swank-exts" - one for the core runtime and one for contribs.

### DIFF
--- a/NEWS
+++ b/NEWS
@@ -1,5 +1,7 @@
 * SLIME News                                -*- mode: outline; coding: utf-8 -*-
 * 2.29 (unreleased)
+** Core
+*** Loading the system "swank" with ASDF produces compilation artifacts
 ** ABCL
 *** Fix missing source position from string buffer location 
 * 2.28 (January 2023)


dkochmanski:
(of course the unspoken system that does not dump images is ecl ^_^)

dkochmanski:
I've removed fully the kludge that prevented reloading of swank if it is in the different location than the version already loaded in the image. If the user has swank in a custom location and tries to load it, then it is either a deliberate action or broken environment. I'd rather punish users with broken environment than users who know what they are doing (i.e by specifying swank as a dependency).

dkochmanski:
Somewhat related to #83 and #80 (more to the latter).

dkochmanski:
Merging as is now.

c.f https://irclog.tymoon.eu/libera/%23slime?around=1678778419#1678778419

PR #767, opened 2023-05-04
Fix token-symbol-thoroughly case conversion, close #766

### LINKED ISSUE #766: tokenize-symbol-thoroughly is incorrect under *print-case*=:invert
```
> (swank::tokenize-symbol-thoroughly "cl-user:Test")
"tEST"
"CL-USER"
nil
```
This is incorrect, `symbol-name` of `cl-user:Test` is `Test`. `*print-case*=:invert` only invert the case if *all* characters are upper or lower case.

I'm looking at a way to fix `tokenize-symbol-thoroughly`. We can add the above test, or I wonder can we reuse the Lisp reader/printer and do something like
```
(defun convert-case (string)
  "Convert external representation of STRING into internal representation
for `package-name's and `symbol-name's."
  (let ((*print-case* :upcase))
    (format nil "~a" (make-symbol string))))
(defun unconvert-case (string)
  (let ((*print-case* :downcase))
    (format nil "~a" (make-symbol string))))
```
Maybe we can even replace the current logic that handles escape character and delegate to the Lisp reader/printer. Is this a good idea?

kchanqvq:
Fix for `*print-case*=:invert`.

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -1542,37 +1542,53 @@ considered to represent a symbol internal to some current package.)"
 
 (defun tokenize-symbol-thoroughly (string)
   "This version of TOKENIZE-SYMBOL handles escape characters."
-  (let ((package nil)
-        (token (make-array (length string) :element-type 'character
-                           :fill-pointer 0))
-        (backslash nil)
-        (vertical nil)
-        (internp nil))
-    (loop for char across string do
-          (cond
-            (backslash
-             (vector-push-extend char token)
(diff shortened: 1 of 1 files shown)


stassats:
```
(values
 (read-from-string "|abc|")
 (swank::tokenize-symbol-thoroughly "|abc|"))
=>
|abc|
"ABC"
```

kchanqvq:
fixed, please take another look, thanks

PR #770, opened 2023-05-26
Revert commit that broke vlime's output buffer

equwal:
While this might be more responsive, it broke vlime. I left the buffer size increase in though.

Test case was:
start vlime
load (defun foo ()) ; FOO
load (defun foo ()) again
did vlime output?: ; FOO

### DIFF
--- a/contrib/slime-repl.el
+++ b/contrib/slime-repl.el
@@ -1729,10 +1729,8 @@ expansion will be added to the REPL's history.)"
 
 (defun slime-repl-event-hook-function (event)
   (slime-dcase event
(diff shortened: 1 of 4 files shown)


equwal:
https://github.com/vlime/vlime/issues/75

stassats:
What does vlime have to do with slime?

equwal:
Vlime is a non-emacs swank client.

stassats:
Then whatever problems vlime is having are vlime problems, not slime problems. Slime is not a library with a public API.

PR #771, opened 2023-06-05
Align compilation artifact paths in swank-loader with ASDF

easye:
When loaded via ASDF, we align the `swank-loader:*source-directory*` and `swank-loader:*fasl-directory*` special variables to the appropriate values.  This restores the ability to launch a Swank server from a Lisp image which has loaded SLIME via ASDF.

See
<https://github.com/slime/slime/commit/dd179f4a0c6874fe0e49fb6c460e9e52a5f58833> and <https://github.com/slime/slime/pull/760>.

-- 

In debugging ABCL/SLIME, I often start ABCL under a JVM debugger, using ASDF to locate and load SLIME.  Without these changes,  @dkochmanski's recent work on `swank.asd` broke this usage.

### DIFF
--- a/swank.asd
+++ b/swank.asd
@@ -20,9 +20,9 @@
 
 (asdf:defsystem "swank"
   :perform (load-op :after (o c)
-              (set (intern '*source-directory* 'swank-loader)
+              (set (intern '*SOURCE-DIRECTORY* 'swank-loader)
                    (asdf:system-source-directory :swank))
-              (set (intern '*fasl-directory* 'swank-loader)
+              (set (intern '*FASL-DIRECTORY* 'swank-loader)
                    (asdf:apply-output-translations (asdf:system-source-directory :swank)))
               (uiop:symbol-call :swank :before-init
                  (uiop:symbol-call :swank-loader :slime-version-string)
--- a/swank.asd
+++ b/swank.asd
@@ -9,17 +9,28 @@
 
 ;; Usage:
 ;;
-;;   (asdf:load-system "swank")
-;;   (swank:create-swank-server PORT) => ACTUAL-PORT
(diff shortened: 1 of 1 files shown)


swank.asd
```diff
@@ -9,17 +9,28 @@
...
 ;;
 ;; This code has been placed in the Public Domain.  All warranties
 ;; are disclaimed.
 
 (asdf:defsystem "swank"
+  :perform (load-op :after (o c)
+              (set (intern '*source-directory* 'swank-loader)
+                   (asdf:system-source-directory :swank))
+              (set (intern '*fasl-directory* 'swank-loader)
+                   (asdf:apply-output-translations (asdf:system-source-directory :swank)))
```
dkochmanski, line 26: these calls to intern break the build, because the argument should be a string (like in calls below(!)). After fixing that all works fine with ECL.

```diff
modified   swank.asd
@@ -20,9 +20,9 @@
 
 (asdf:defsystem "swank"
   :perform (load-op :after (o c)
-              (set (intern '*source-directory* 'swank-loader)
+              (set (intern "*SOURCE-DIRECTORY*" 'swank-loader)
                    (asdf:system-source-directory :swank))
-              (set (intern '*fasl-directory* 'swank-loader)
+              (set (intern "*FASL-DIRECTORY*" 'swank-loader)
                    (asdf:apply-output-translations (asdf:system-source-directory :swank)))
               (uiop:symbol-call :swank :before-init
                  (uiop:symbol-call :swank-loader :slime-version-string)
```

dkochmanski (changes requested):

swank.asd
easye: Thanks for the review!

PR #774, opened 2023-06-08
Simple fixes - like typos.

phmarek:


phmarek:
When deserializing data via non-default means -- eg. `YASON` may (in some cases) return non-simple strings.

phmarek:
Also, I'm using that function to provide completion in a web-interface, and depending on how the string gets built it might not be simple.

As it's a `swank` internal requirement that the string must be simple, I guess this interface is the right place to ensure it.

stassats:
I wouldn't say that swank provides any interface beyond what slime uses.

PR #787, opened 2023-08-04
Update ~/.slime to use $XDG_DATA_HOME

ron-at-swgy:
This commit updates the `~/.slime` references in `Makefile` and `swank-loader.lisp` to check for XDG_DATA_HOME. If the environment variable is set, it uses $XDG_DATA_HOME/slime as the default fasl directory.

This change is necessary as there is no way to avoid creation of "~/.slime" when loading a system that requires Swank. My personal example is trying to port a Clack application to OpenBSD. The OpenBSD build system does not allow writing to home directories as part of the build process.

Prior to these changes, I tried using asdf:defsystem-depends-on, asdf:around-compile, and use of a secondary system to only load swank.


ron-at-swgy:
Relates to #610 https://github.com/slime/slime/issues/610

stassats:
Can it be made optional?

ron-at-swgy:
Using the environment variable alone seems to be the least intrusive means to change the behavior. 

How would you like it to be made optional? There is no way to pass command-line arguments or any other switches during (for instance) a quickload installation of swank. A separate environment variable could be used, or a sentinel file, but these methods all seem to add more cludge that necessary.

Please advise

ron-at-swgy:
The use of XDG_DATA_HOME comes from the [Free Desktop specification](https://specifications.freedesktop.org/basedir-spec/basedir-spec-latest.html#basics).

stassats:
I don't have XDG_DATA_HOME or Free Desktop on macOS, so having different locations on the two systems I use at the same time is going to be inconvenient.

ron-at-swgy:
I don't have XDG_DATA_HOME on OpenBSD either (it would depend on the window manager choice). My motivation for the change is to have control over the location of the slime directory during installation. Would using a "SLIME_DATA_DIRECTORY" environment variable do the trick? That would mean the current behavior only changes if someone chooses to change it. One could then set SLIME_DATA_DIRECTORY=XDG_DATA_HOME if that's the behavior they want.

Different environment variable?

ron-at-swgy:
Closed due to inactivity

PR #788, opened 2023-08-27
Ability to change buffer action for show-source-location

bmiww:
Added a variable to be able to overwrite how `slime-show-source-location` handles the newly opened buffer.
Making changes to `display-buffer-alist` for this use case seems to be overly involved.

Example configuration (I like My buffers in new frames):
`(setq slime-show-source-location--display-buffer-action '(display-buffer-pop-up-frame . '()))`

I see that there are also some var customisation groups, could not decide which this variable would fit in with, if any at all:
slime-debugger or slime-ui.

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -5719,12 +5719,14 @@ This is 0 if START and END at the same line."
         (t
          (slime-show-source-location source-location t nil))))))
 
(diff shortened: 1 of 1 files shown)


bmiww:
Nevermind, finally figured out how to do this kind of thing completely with just display-buffer-alist.

PR #793, opened 2023-09-27
clasp: new xref implementation

Bike:
Conditionalized like SBCL's so it works with old Clasps. Basically just calls new EXT interface functions of the same names.

### DIFF
--- a/swank/clasp.lisp
+++ b/swank/clasp.lisp
@@ -28,6 +28,13 @@
   (when (find-symbol "TEMPORARY-DIRECTORY" "EXT")
     (pushnew :temporary-directory *features*)))
 
+;;; Compatibility tests
+
+(eval-when (:compile-toplevel :load-toplevel :execute)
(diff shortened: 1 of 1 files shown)


stassats:
How prevalent are old clasp versions? And how old is old?

Bike:
Not very prevalent. I just meant that Clasp builds from before this xref addition will still work. Back to at least Clasp 2.1 from January.

stassats:
I'm just assuming that people who use clasp use the latest version. Doing it like that is fine too but, as usual, the less code the better.

Bike:
Yeah, I'm just trying to keep things more stable moving forward. I can delete the conditional code if you'd prefer.

stassats:
I'd prefer to do whatever the clasp users would prefer to do.

Bike:
OK, can we stick with this then?

Bike:
Thanks

PR #806, opened 2023-12-28
fix HANDLER-CASE indentation

galdor:
Indentation of HANDLER-CASE error clauses is currently confusing when the condition type specifier is a list spanning multiple lines. The most common case is a OR type specification too long for a single line:

    (handler-case
        nil
      ((or foo-error bar-error
        my-very-long-condition-name)
        (condition)
        nil)
      (baz-error
        (condition)
        nil)
      (error (condition)
        nil)
      (:no-error
        (a b c)
        nil))

The OR form is incorrectly indented, and the second element (the list containing the condition variable name) is indented at the same column as the following form(s).

This patch modifies indentation so that the type specification list is correctly indented. It adds one indentation level to the condition variable list to make it easier to distinguish from subsequent forms.

    (handler-case
        nil
      ((or foo-error bar-error
           my-very-long-condition-name)
          (condition)
        nil)
      (baz-error
          (condition)
        nil)
      (error (condition)
        nil)
      (:no-error
          (a b c)
        nil))


stassats:
Is it consistent with the other places where type specifiers are indented?

galdor:
> Is it consistent with the other places where type specifiers are indented?

Do you have examples in mind? Regarding the indentation in the `OR` form of the type specifier, both `CHECK-TYPE` and `DECLARE` will indent it the same way if this is what you mean.

stassats:
```
(typep x '(or list
           integer))
```

doesn't.

galdor:
Correct, but `TYPEP` (and any form using a type specifier which is evaluated during execution) is already inconsistent with `DECLARE`, `CHECK-TYPE` & co. I'm not even sure SLIME indentation can handle it.

Thank you for merging this!

PR #823, opened 2024-04-23
Add mode: REPL history commands (M-p etc.) insert, not replace.

slburson:
The REPL history commands (`M-p`/`M-n`) do not work the way I prefer.  After I've evaluated a top-level form, I frequently want to wrap it in some more calls and evaluate it again.  (Yes, I know about the `*` variables, but I don't always like to use them; the expression I want to wrap and re-evaluate may not be one of the last three I entered, or maybe I just don't want to deal with the context sensitivity.)  What I like to be able to do is to type the new calls first, then pull up the desired form from the history, such that it's inserted at point, then hit Enter.  I use Paredit, so I shouldn't even need to type any close-parens.  For example, given
```
CL-USER> (foo)
"Value of foo"
CL-USER>
```

I want to be able to type `( b a r Space M-p Enter` to evaluate `(bar (foo))`.

The existing commands do not support this usage; if I type anything before doing `M-p`, that is automatically taken as a search string.  I can disable that by doing `C-Space C-u M-p`, but then I lose the close paren(s) _following_ point that Paredit inserted for me; I have to put them back manually, which in my configuration requires typing `C-q` before each one.

The changes in this PR are to support my preferred usage model.  They are conditioned on a new custom variable, `slime-repl-history-recall-mode`, which defaults to `replace` so that the standard behavior is unchanged.  When it's set to `insert`, though, automatic searching is turned off, and recalled history elements are inserted at point without deleting any of the user's text.  (Of course one can still use `M-r` to search.)

### DIFF
--- a/contrib/slime-repl.el
+++ b/contrib/slime-repl.el
@@ -55,6 +55,19 @@ maintain."
   :type '(boolean)
   :group 'slime-repl)
 
+(defcustom slime-repl-history-recall-mode 'replace
+  "When REPLACE, a recalled history item replaces the current input.
+   When INSERT, a recalled history item is inserted at point, making
+   it much easier to nest a previous input inside the current one,
+   particularly if you are using Paredit. INSERT also turns off the
+   behavior whereby M-p (slime-repl-previous-input) and
+   M-n (slime-repl-next-input), if you have already typed some characters,
+   automatically use them as a history search string; if you want to
+   search the history, use M-r (slime-repl-previous-matching-input)."
+  :type '(radio (const replace)
+		(const insert))
+  :group 'slime-repl)
+
 (defcustom slime-repl-auto-right-margin nil
   "When T we bind CL:*PRINT-RIGHT-MARGIN* to the width of the
 current repl's (as per slime-output-buffer) window."
@@ -966,7 +979,7 @@ Empty strings and duplicates are ignored."
       (push string slime-repl-input-history))))
 
 ;; These two vars contain the state of the last history search.  We
-;; only use them if `last-command' was 'slime-repl-history-replace,
+;; only use them if `last-command' was 'slime-repl-history-recall,
 ;; otherwise we reinitialize them.
 
 (defvar slime-repl-input-history-position -1
@@ -975,8 +988,10 @@ Empty strings and duplicates are ignored."
 (defvar slime-repl-history-pattern nil
   "The regexp most recently used for finding input history.")
 
-(defun slime-repl-history-replace (direction &optional regexp)
-  "Replace the current input with the next line in DIRECTION.
+(defun slime-repl-history-recall (direction &optional regexp)
+  "Recall the next history entry in DIRECTION.
+Replace the current input with it if SLIME-REPL-HISTORY-RECALL-MODE
+is REPLACE, or just insert it at point if this mode is INSERT.
 DIRECTION is 'forward' or 'backward' (in the history list).
 If REGEXP is non-nil, only lines matching REGEXP are considered."
   (setq slime-repl-history-pattern regexp)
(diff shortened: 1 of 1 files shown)


contrib/slime-repl.el
```diff
@@ -975,8 +988,10 @@ Empty strings and duplicates are ignored."
 (defvar slime-repl-history-pattern nil
   "The regexp most recently used for finding input history.")
 
-(defun slime-repl-history-replace (direction &optional regexp)
-  "Replace the current input with the next line in DIRECTION.
+(defun slime-repl-history-recall (direction &optional regexp)
+  "Recall the next history entry in DIRECTION.
+Replace the current input with it if SLIME-REPL-HISTORY-RECALL-MODE
+is REPLACE, or just insert it at point if this mode is INSERT.
```
slburson, line 994: I renamed this since it no longer necessarily replaces.

stassats:
That is a somewhat confusing option to have, can you achieve the behavior you want by adding an advice for slime-repl-replace-input?

slburson:
For my own use, I don't really care whether you merge the PR; I can run my forked version.  But I expect other Paredit users will appreciate this option.  (If you do merge it, I will add a note [here](https://www.emacswiki.org/emacs/ParEdit#h5o-4) to help people discover the feature.

Also, if you try the `insert` mode, I think you'll find that what it does is very simple.  It's just like `C-y`/`M-y`.  It's also how Franz Allegro ELI has always worked (I used it for many years).

luismbo:
FWIW, this has been proposed before: https://github.com/slime/slime/pull/184

I wonder if this feature could be offered as independent interactive commands?

slburson:
I've done some refactoring which I hope will make this code both more flexible and easier for users to understand.  Instead of a single customization variable, `slime-repl-history-recall-mode`, there are now two:

- `slime-repl-history-autosearch-mode` controls what happens when `M-p` is typed when the input line is nonempty.  If it's true, the contents are used as a search string as before; if false, this behavior is disabled.
- `slime-repl-paredit-compatibility` makes a few tweaks to make things work more nicely for users of ParEdit and other balanced-delimiter-editing packages.

While I was at this, I couldn't resist making it so that walking off either end of the history (unless you have history wrapping turned on) puts your input line back to the state it was in before you first typed `M-p`.  When autosearch is on, this requires re-inserting the original text; in ParEdit compatibility mode, it even puts the trailing parens back, with point in the right place.

The other thing ParEdit compatibility mode does is to make `C-c C-u` kill the entire input, not just from the prompt to point, since point is rarely at the end in ParEdit mode.  There may well be more such tweaks that would be desirable that I haven't noticed yet.

I did make one incompatible change to existing behavior.  If you type some text and then hit `M-r`, the existing text is retained, and the history entry is inserted at point.  Previously, the existing text was simply discarded in this case; I don't expect anybody will be especially attached to this behavior, but I could be wrong :smile_cat:

I have made some effort to test this code, but I'm not sure it's ready for merging yet.

slburson:
Oh, thanks, @luismbo.  I'm glad to have @rpgoldman also agitating for this feature.  Of course there are multiple ways to implement it; if it's considered preferable to do it with key rebinding rather than customization variables, I can redo it.  But it seems to me that customization variables will be easier for users to find.

slburson:
I've been using this for months now, and it seems to work fine.  I actually think it improves REPL history behavior for everyone, even those who don't set the new customization variables.  This is because if you hit `M-p` a few times, and then `M-n` an equal number of times, it puts your input line back in the state it was in before the first `M-p`.  In this it duplicates the familiar behavior of `bash` and `tcsh`.

I have been using it with `slime-repl-history-autosearch-mode` set to false, and `slime-repl-paredit-compatibility` set to true.  I have tested the other three states of these two variables, but I haven't lived with those other settings long enough to be certain they never do anything infelicitous.  So before this gets merged, it would be good to have experience reports from users who prefer the other states.  But I think the PR is ready for such beta testing.

I definitely recommend that @rpgoldman give it a try.

lispstudent:
I saw your comment :)

I meant to point to [mmontone reply](https://www.reddit.com/r/lisp/comments/1nduk3f/github_mmontoneslimestar_slime_configuration_with/ndm9lp3/):

PR #832, opened 2024-07-05
Fix top-level defvar detection in slime-eval-defun

hraban:
Using ‘(string-match "^..." ..)’ would match against the beginning of every line, not just the start of the string. This trips over docstrings which have a (defvar..), e.g. in example usage.

Compare the old and new implementation on:

```
(defun foo (x)
  "Foo something

Example:

(defvar y 10)
(foo y)
=> 100
"
  (* x x))
```

I think ‘string-prefix-p’ goes back at least to 24.1, if even earlier support is needed maybe we should use a different regex instead?

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -4118,7 +4118,7 @@ inserted in the current buffer."
 Use `slime-re-evaluate-defvar' if the from starts with '(defvar'"
   (interactive)
   (let ((form (slime-defun-at-point)))
-    (cond ((string-match "^(defvar " form)
+    (cond ((string-prefix-p "(defvar " form t)
            (slime-re-evaluate-defvar form))
           (t
            (slime-interactive-eval form)))))


stassats:
Shouldn't it be case-insensitive?

hraban:
You're totally right, great catch. I've changed the PR. I used `(rx)` here because the string notation is so weird:

    "\\`(defvar"

That it seemed genuinely less confusing to just have `(rx string-start ...)` even if someone doesn't know rx. But of course feel free to just put the string version instead.

stassats:
Can you try your first approach but with a third argument to string-prefix-p?

hraban:
> Can you try your first approach but with a third argument to string-prefix-p?

haha I thought you meant you wanted to keep regex. ok pushed.

PR #838, opened 2024-07-31
Fix slime-modeline-string

mavbozo:
Fixes slime/slime/issues/836

In emacs version < 28, slime-modeline-string execution throws (void-function string-replace). string-replace exists in emacs version > 28.

This commit changes string-replace usage with replace-regexp-in-string and also add tests for slime-modeline-string function.

### DIFF
--- a/slime-tests.el
+++ b/slime-tests.el
@@ -1328,6 +1328,26 @@ Reconnect afterwards."
                               (not (member hook slime-connected-hook)))
                             5))))
 
+(def-slime-test slime-modeline-string-test
+  (expected current-connection buffer-connection
+   connection-name current-package modeline-state-string)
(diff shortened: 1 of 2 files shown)


stassats:
Are the "t t" arguments needed?

mavbozo:
> Are the "t t" arguments needed?

Yes, I want the result of applying `replace-regexp-in-string` to be the same as the previous `string-replace`.  The first `t` for `fixedcase` arg means no case conversion. The second `t` for `literal` arg means no special interpretation for `\` in the replacement string.

stassats:
Does % have a case? Does "%" have \?

mavbozo:
> Does % have a case? Does "%%" have \?

No. I don't have to use the additional args for that usage.

I've removed the additional args from the commit.

PR #846, opened 2024-10-18
slime-redirect-inferior-output is useful and still in use

gutron:
Revert two recent CLs that removed useful -- and still in-use -- functionality.


stassats:
The output is always redirected.

PR #848, opened 2024-11-04
Handle carriage in output

mmontone:
This patch makes carriage character and friends to be handled in slime-repl output, when `*slime-repl-carriage-motion*` is enabled. 
Code is based of what `comint` does to processes output.
This is useful for displaying things like progress bars, that use carriage return to update the display, from the SLIME repl.  

progressons


mmontone:
There is an issue with highlighting and presentations and the handling of characters that result in the deletion of previously output text. That's why slime-propertize-region is invoked with arguments or not depending on the value of slime-repl-carriage-motion.

PR #850, opened 2024-11-22
Customize offsets when showing maximum output or recentering SLIME REPL

vhuynhle:
Before this commit, the values are hard-coded to -1, which are sometimes too small and the output is cut off at the bottom of the screen.

### DIFF
--- a/contrib/slime-repl.el
+++ b/contrib/slime-repl.el
@@ -61,6 +61,22 @@ current repl's (as per slime-output-buffer) window."
   :type '(boolean)
   :group 'slime-repl)
 
+(defcustom slime-repl-show-maximum-output-recenter-arg -1
+  "Offset from the bottom of the screen when showing the maximum
+amount of output in the SLIME REPL.
+If the number is negative (recommended setting),
+it counts up from the bottom of the window."
+  :type '(integer)
+  :group 'slime-repl)
+
+(defcustom slime-repl-return-recenter-arg -1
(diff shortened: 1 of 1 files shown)


stassats:
When does this happen? Why does it need to be a variable if -1 is wrong? How do you figure out the right value?

vhuynhle:
1. When does this happen? 
I have observed this on Fedora and also Windows. Here is an example on Fedora with the text slightly cut-off after sending an expression from the Lisp buffer with C-c C-j:

example-fedora

I will try to reproduce later on Windows.

In addition, I want to have some "breathing" room at the end of the buffer, even if the text is not cut off. Related question on [StackOverflow](https://stackoverflow.com/questions/71654967/portacle-lisp-slime-repl-sbcl-prompt-keep-auto-scrolling-back)

3. Why does it need to be a variable if -1 is wrong?
-1 is only "wrong" on my machine as far as I can tell. I keep the default value the same so that there is no change needed by other users (if their setups work).

4. How do you figure out the right value?
I play with the value until I have a comfortable space at the end of the buffer.
I have settled for -5 or even -10.

3, 4: Personally, I'm fine with hard-coding the values at -5, but others may not like this exact value. That why I make them customizable with the default values the same as before.

vhuynhle:
Here's what look like with offset  = 5:

example-offset-5

stassats:
That sounds like a bug in emacs. I can't achieve that even when scrolling with a mouse.

PR #865, opened 2025-03-31
Move the use of *cache-sourcecode* to avoid calling #'truename

K1D77A:
Hey, I found that when I was connected to my remote image, conditions related to files would be signalled. From the stack trace I could see it was related to the source file caching so I set the *cache-sourcecode* when starting the swank server on my remote image and I still had the same problems. After a little inspection I moved the use of *cache-sourcecode* to avoid the call to #'truename and the problem went away. 
This will maintain the previous functionality.


stassats:
Have you tried the current git HEAD?

K1D77A:
I'll give it a try and report back

K1D77A:
Yep, latest version sorted it. Thanks for doing that.

PR #871, opened 2025-04-27
add command slime-presentation-menu-at-point

fstamour:
to be able to open a presentation's menu without having to click using the mouse, which might not even be possible when using emacs in the terminal.


stassats:
Isn't there something in emacs for that? And there's xterm-mouse-mode anyway.

fstamour:
> Isn't there something in emacs for that? 

if there's something, I don't know about it.

> And there's xterm-mouse-mode anyway.

I'll try that out

stassats:
M-x context-menu-open

stassats:
You can add some function for yourself to trigger mouse events.

PR #872, opened 2025-06-02
slime-indentation.el: use the canonical package name for comparisons

melisgl:
Previously, common-lisp-get-indentation did not use the indentation associated with some symbol name in the current package reliably because it compared slime-current-package (e.g :pgkname, #:pkgname, or "pkgname") to PKGNAME. Without an exact package name match, it heuristically chose the indentation setting associated with the most packages.

This change looks up the canonical package name on the Common Lisp side, so the above heuristic need not trigger in most cases. If Slime is not connected, then the canonical package name is guessed heuristically, so that any leftover data in common-lisp-system-indentation may still be matched.


luismbo:
Thanks!

PR #876, opened 2025-08-07
Ignore errors in find-package during completion

dieggsy:
When typing `excl::` in Allegro Common Lisp an implementation-specific error occurred.

This was the backtrace
```
The parent of package #<The KEYWORD package> does not exist.
   [Condition of type SIMPLE-ERROR]

Restarts:
 0: [*ABORT] Return to SLIME's top level.
 1: [ABORT] Return to Top Level (an "abort" restart).
 2: [ABORT] Abort entirely from this (lisp) process.

Backtrace:
  0: (ERROR "The parent of package ~a does not exist." #<The KEYWORD package>)
  1: (PACKAGE-PARENT #<The KEYWORD package>)
  2: (RELATIVE-PACKAGE-NAME-TO-PACKAGE "..ENVIRONMENT..")
  3: (SWANK::SYMBOL-CLASSIFICATION-STRING EXCL::..ENVIRONMENT..)
  4: (SWANK::FORMAT-COMPLETION-SET ..)
```

### DIFF
--- a/swank.lisp
+++ b/swank.lisp
@@ -2729,7 +2729,7 @@ boundp fboundp generic-function class macro special-operator package accessor"
       (when (find-class symbol nil)   (flip #\c) )
       (when (macro-function symbol)   (flip #\m))
       (when (special-operator-p symbol) (flip #\s))
-      (when (find-package symbol)       (flip #\p))
+      (when (ignore-errors (find-package symbol)) (flip #\p))
       (when (structure-accessor-p symbol) (flip #\a))
       result)))
 


stassats:
That seems like a bug in ACL, why would find-package signal an error.

melisgl:
On AllegroCL, FIND-PACKAGE will signal an error if a relative package name has too many leading dots. On CMUCL, (FIND-PACKAGE "..") fails.

stassats:
I'm still of the opinion that these are their bugs.

melisgl:
I just wanted to some more context / data. I'm not arguing either way. AllegroCL and CMUCL extended the package system in a slightly incompatible way.

PR #879, opened 2025-08-19
Changes for connections from the web browser

dkochmanski:
This set of changes introduces a new connection type `:ASYNC` that relies on Gray Stream interfaces. Moreover SLIME-NET-SEND anticipates that the connection may be associated with a websocket, and if that's the case we use a specialized function to send data.

With these changes I've managed to successfully connect from an image running inside the web browser to a shim server bridging SLIME and websockets.

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -2000,7 +2000,7 @@ Return nil if there's no process object for the connection."
   (let ((c (or connection (slime-connection))))
     (cl-ecase (slime-communication-style c)
       ((:fd-handler nil) t)
-      ((:spawn :sigio) nil))))
+      ((:spawn :sigio :async) nil))))
 
 (defvar slime-inhibit-pipelining t
   "*If true, don't send background requests if Lisp is already busy.")
--- a/swank.lisp
+++ b/swank.lisp
@@ -171,7 +171,7 @@ Backend code should treat the connection structure as opaque.")
   ;; The list of packages represented in the cache:
   (indentation-cache-packages '())
   ;; The communication style used.
-  (communication-style nil :type (member nil :spawn :sigio :fd-handler))
+  (communication-style nil :type (member nil :async :spawn :sigio :fd-handler))
   )
 
 (defun print-connection (conn stream depth)
@@ -213,11 +213,16 @@ Backend code should treat the connection structure as opaque.")
   (active-threads '() :type list)
   )
 
+(defstruct (asynchronous-connection (:include singlethreaded-connection)
+                                    (:conc-name aconn.)))
+
 (defvar *emacs-connection* nil
   "The connection to Emacs currently in use.")
 
 (defun make-connection (socket stream style)
   (let ((conn (funcall (ecase style
+                         (:async
+                          #'make-asynchronous-connection)
                          (:spawn 
                           #'make-multithreaded-connection)
                          ((:sigio nil :fd-handler)
@@ -699,12 +704,15 @@ e.g.: (restart-loop (http-request url) (use-value (new) (setq url new)))"
              (note () (add-server socket port (current-thread)))
              (serve-loop () (note) (loop do (serve) while dont-close)))
       (ecase style
-        (:spawn (initialize-multiprocessing
-                 (lambda ()
-                   (if (or dont-close
-                           *main-thread-used*)
-                       (spawn #'serve-loop :name (format nil "Swank ~s" port))
-                       (serve-loop)))))
+        (:async
+         (error "SETUP-SERVER can't run with a ~s connection." style))
+        (:spawn
+         (initialize-multiprocessing
+          (lambda ()
+            (if (or dont-close
+                    *main-thread-used*)
+                (spawn #'serve-loop :name (format nil "Swank ~s" port))
+                (serve-loop)))))
         ((:fd-handler :sigio)
          (note)
          (add-fd-handler socket #'serve))
@@ -773,6 +781,8 @@ if the file doesn't exist; otherwise the first line of the file."
 (defun serve-requests (connection)
   "Read and process all requests on connections."
   (etypecase connection
+    (asynchronous-connection
+     (error "SERVE-REQUSTS not implemented for :ASYNC connection."))
     (multithreaded-connection
      (spawn-threads-for-connection connection))
     (singlethreaded-connection
@@ -783,6 +793,9 @@ if the file doesn't exist; otherwise the first line of the file."
 
 (defun stop-serving-requests (connection)
   (etypecase connection
+    (asynchronous-connection
+     ;; The act of closing the stream terminates the :ASYNC connection.
+     t)
     (multithreaded-connection
      (cleanup-connection-threads connection))
     (singlethreaded-connection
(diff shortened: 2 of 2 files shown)


swank.lisp
```diff
@@ -1073,9 +1086,32 @@ event was found."
       (etypecase c
         (multithreaded-connection
          (receive-if (lambda (e) (event-match-p e pattern)) timeout))
+        (asynchronous-connection
+         (wait-for-event/async-read c pattern timeout))
         (singlethreaded-connection
          (wait-for-event/event-loop c pattern timeout))))))
 
+;;; This function is similar to WAIT-FOR-EVENT/EVENT-LOOP, but instead of
```
stassats, line 1094: I don't understand what's going on here at all.

dkochmanski: This function is similar to wait-for-event/event-loop, but instead of calling backend's `wait-for-input` (that works only on implementaiton-specific network streams) we see whether there is some input enqueued with `listen` called on the stream, and if there is we call `decode-message` that will read from the stream - it is assumed that if there is input available in the stream then it is a complete message (or, if the timeout is nil, that reading from the stream is a blocking operation).

I have two clients for this interface that are not suitable for inclusion in SWANK, because they depend on other libraries, but the gist is that the user can pass their own stream with implemented appropriate methods and SWANK will be able to work with that. Here are clients (both depend on the file slug.lisp):

https://turtleware.eu/static/paste/a62429c3-slug.lisp
https://turtleware.eu/static/paste/ecd17baa-frig.lisp
https://turtleware.eu/static/paste/72d734fe-wank.lisp

The frig client can be used from sbcl pulling appropriate websocket client library from quicklisp.

stassats: Can't wait-for-input be made to do what listen does?

slime.el
```diff
@@ -1489,9 +1489,12 @@ EVAL'd by Lisp."
                    (concat (slime-prin1-to-string sexp) "\n")
                    'utf-8-unix))
          (string (concat (slime-net-encode-length (length payload))
-                         payload)))
+                         payload))
+         (websocket (process-get proc :websocket)))
     (slime-log-event sexp)
-    (process-send-string proc string)))
+    (if websocket
+        (websocket-send-text websocket string)
```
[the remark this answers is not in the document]
dkochmanski: By adapter I mean a glue between SLIME and Websockets. The process that is passed to SLIME is an ordinary TCP, but Websockets require their messages to be formatted (headers and such), that's why we need to pass data through a different function.

swank.lisp
dkochmanski: Multiple backends to SWANK have different implementations of wait-for-input that depend on different mechanisms (most notably at least three implementations depend on `serve-event` that is not compatible with gray streams).

Moreover this solution does not rely on any implementation-specific code, so there is no need to making it an interface that is later filled by the backend.

dkochmanski:
I'll look into CCL tests not passing tomorrow. Thanks for reviewing these changes.


EDIT: CCL failure was caused by not having all gray stream protocol operators implemented and exported. I've withdrawn the part that pulls symbols to swank's gray package because of that and left only essential parts.

swank.lisp
stassats: But it makes sense to me, so that's the only variation I'm going to accept.

dkochmanski: Please reconsider. Is there other way to write a client like this:

https://turtleware.eu/static/paste/a62429c3-slug.lisp
https://turtleware.eu/static/paste/ecd17baa-frig.lisp

that will allow me to test the connection with websockets from the "ordinary" host (that is not in the browser) across supported implementations? Please advise.

Moreover using `wait-for-input` won't be useful here, because it is the methods `stream-read-byte` that yields control when there is no input, so pooling over listen on empty stream, in this environment, won't progress anything.

If you insist on doing it this way I will also change the stream-listen method to yield control on this call, and when the thread is scheduled again recheck whether the input is available, although I don't think that it is a good approach.

In that case, could you suggest a solution for running a portable websockets client to connect from swank to slime?

dkochmanski:
I've submitted a minimal set of changes to slime.el and for ecl backend (as requested) and they were merged some time ago. The only remaining change is to introduce `:async` communication style.

This style is meant for swank communicating through a stream, where the user implements the gray protocol that handles the communication without assumptions that there is an underlying file handler (many backends make that assumption).

It is called `:async`, because the control loop that is usually started by `server-requests` is not invoked from swank, but rather the client serves requests at its own discretion (most notably, from a callback that may signal that there is a new message sent from slime).

Why singlethreaded-connection or multithreaded-connection are not enough:

- multithreaded-connection assumes threads, while in case of :async, it is at the implementor discretion whether the connection runs in a single or in multiple threads

- singlethreaded-connection style implies `slime-use-sigint-for-interrupt`, but most notably `wait-for-event/event-loop` is structured in a way, that it never returns when in the debugger context (because `wait-for-input` is called even when timeout is set to `nil`)

```
(defun wait-for-event/event-loop (connection pattern timeout)
  (assert (or (not timeout) (eq timeout t)))
  (loop 
   (check-slime-interrupts)
   (let ((event (poll-for-event connection pattern)))
     (when event (return (car event))))
   (let ((events-enqueued (sconn.events-enqueued connection))
         (ready (wait-for-input (list (current-socket-io)) timeout)))
     (cond ((and timeout (not ready))
            (return (values nil t)))
           ((or (/= events-enqueued (sconn.events-enqueued connection))
                (eq ready :interrupt))
            ;; rescan event queue, interrupts may enqueue new events 
            )
           (t
            (assert (equal ready (list (current-socket-io))))
            (dispatch-event connection
                            (decode-message (current-socket-io))))))))
```

Contrast this with the variant for async connection, where `listen` is called only when timeout is true (after checking whether there are already available events):

```
(defun wait-for-event/async-read (connection pattern timeout)
  (assert (or (not timeout) (eq timeout t)))
  (loop
    (check-slime-interrupts)
    (let ((event (poll-for-event connection pattern)))
      (when event (return (car event))))
    (let ((events-enqueued (aconn.events-enqueued connection)))
      (cond
        ;; New events pushed or executed an interrupt. Loop over.
        ((or (/= events-enqueued (aconn.events-enqueued connection))
             (check-slime-interrupts))
         nil)
        ;; No input available and we are not about to be waiting.
        ((and timeout (not (listen (current-socket-io))))
         (return (values nil t)))
        ;; Process the next message as soon as it is available.
        ((dispatch-event connection
                         (decode-message (current-socket-io))))))))
```

This second version ensures, that when there is no timeout, we will proceed straight to `decode-message` that will block until there is new input. Why it matters in WASM port:

When a condition is signaled, we enter sldb-loop to allow debugging the condition from emacs. Understandably, to preserve context, it doesn't return until we are done with the debugger. So `wait-for-event` is called with `timeout` set to `nil`.

In my port, `stream-listen` is implemented by checking whether there is data in the buffer, while reading from the stream (a blocking operation), will yield control to javascript, that will invoke the callback when event occurs, effectively populating the buffer, and returning control to `stream-read` (called by `decode-message`).

So if `wait-for-input` is called when `timeout` is nil, it will wait forever checking whether the stream has data (and it won't, because we didn't return control to the main process by yielding).


This set of changes does not impact existing code paths, so it should be safe to merge. I also think that it is a useful general-purpose solution for communication that levereges gray stream protocol in a backend-independent way.

stassats:
Is there code anywhere that uses this new functionality that I can look at?

dkochmanski:
Yes; there are even two clients, so you can try it from SBCL without building WECL in the browser.

- slime adapter to websockets: https://fossil.turtleware.eu/wecl/file?name=App/lime.el

load this file in emacs and evaluate the form `(lime-net-listen "localhost" 8889)` to start listening for connections

- swank adapter to websockets (portable):
  https://fossil.turtleware.eu/wecl/file?name=App/slug.lisp
  https://fossil.turtleware.eu/wecl/file?name=App/frig.lisp

sbcl --load slug.lisp --load frig.lisp
CL-USER> (EU.TURTLEWARE.WECL/FRIG::test-frig)

This should open a new repl in the emacs session. Note that frig.lisp assumes that quicklisp is present and loads "websocket-driver-client" and "bordeaux-threads" system

- swank adapter to websockets (browser):
  https://fossil.turtleware.eu/wecl/file?name=App/slug.lisp
  https://fossil.turtleware.eu/wecl/file?name=App/wank.lisp

This client runs from the browser, i.e using a page like this:

```
<!doctype html>
<html>
  <head>
    <title>Web Embeddable Common Lisp</title>
    <link rel="stylesheet" href="easy.css" />
    <script type="text/javascript" src="boot.js"></script>
    <script type="text/javascript" src="wecl.js"></script>
    <script type="text/common-lisp" src="slug.lisp"></script>
    <script type="text/common-lisp" src="wank.lisp"></script>
    <script type="text/common-lisp" src="easy.lisp">
      (defvar *connect-button* (make-element "button" :text "Connect"))
      (define-js-callback (connect-to-slug :type :null) ((event :js-ref))
       (wank-connect "localhost" 8889)
       (setf (inner-text *connect-button*) "Crash!"))
      (add-event-listener *connect-button* "click" (js-callback connect-to-slug))
      (append-child [body] *connect-button*)
    </script>
  </head>
  <body>
  </body>
</html>
```

you may try it without cross compiling ecl here: https://turtleware.eu/static/misc/wecl-20250821/slug.html, when you click "Connect", it will try to connect to the server created by lime.el at port 8889 at localhost.

Note that loading the page first time may take some time depending on your internet connection, because wecl.wasm is around 10MB.

dkochmanski:
rebased against current master.

swank.lisp
dkochmanski: all prerequisite changes has been merged already, and this change is the last bit stripped to bare minimum. rationale why it is necessary is in later comment

https://github.com/slime/slime/pull/879#issuecomment-4364652539

stassats:
I've been meaning to review it, but I'm always busy with other stuff, and I don't want to hastily introduce new APIs to be stuck with it forever after.

dkochmanski:
Any update on this? technically I'm also slime dev, so if anything breaks due to this new API I will address the issue when @mentioned.

PR #888, opened 2025-09-15
common-lisp-hyperspec-root path handling

dotemacs:
When `common-lisp-hyperspec-root` is set without the leading `file://`, the call to hyperspec-lookup is dropped silently.

When `common-lisp-hyperspec-root` is set without the trailing `/` (on Unix like OS), the hyperspec-lookup is also dropped silently.

Ideal fix for this would be to use something like this:

```elisp
(defcustom common-lisp-hyperspec-root
  "http://www.lispworks.com/reference/HyperSpec/"
  "The root of the Common Lisp HyperSpec URL.
If you copy the HyperSpec to your local system, set this variable to something like \"/usr/local/doc/HyperSpec/\" and it will automatically be prefixed with \"file://\" and normalized as a proper directory path."
  :type 'string
  :set (lambda (symbol value)
         (let ((processed-value
                (if (and (stringp value)
                         (not (string-match-p "^https?://" value))
                         (not (string-match-p "^file://" value)))
                    (concat "file://"
                            (file-name-as-directory
                             (expand-file-name value)))
                  value)))
                  (set-default symbol processed-value))))
```

But since SLIME is over 23 years old, and hyperspec.el is over 28 years old, there might be some backwards compatibiliy issues.

Because you can't set defcustom values with setq.
(Caveat, you theoretically can, in init.el. **So should this PR be revised and instead the above `defcustom` added instead?**)

This fix tries to treat those edge case potential errors by checking the value of `common-lisp-hyperspec-root` and ensures that it has proper formatting, so that the calls aren't dropped silently.

What do you think? Thanks

### DIFF
--- a/lib/hyperspec.el
+++ b/lib/hyperspec.el
@@ -112,7 +112,9 @@ If you copy the HyperSpec to another location, customize the variable
 `common-lisp-hyperspec-root' to point to that location."
   (interactive (list (common-lisp-hyperspec-read-symbol-name)))
   (let* ((name (common-lisp-hyperspec--strip-cl-package (downcase symbol-name)))
-	 (hyperspec-root-path (file-name-as-directory common-lisp-hyperspec-root)))
+	 (hyperspec-root-path (if (string-suffix-p "/" common-lisp-hyperspec-root)
+                                  common-lisp-hyperspec-root
+                                (concat common-lisp-hyperspec-root "/"))))
     (cl-maplist (lambda (entry)
 		  (browse-url (concat hyperspec-root-path "Body/" (car entry)))
 		  (when (cdr entry)
--- a/lib/hyperspec.el
+++ b/lib/hyperspec.el
@@ -111,13 +111,20 @@ Visit http://www.lispworks.com/reference/HyperSpec/ for more information.
 If you copy the HyperSpec to another location, customize the variable
 `common-lisp-hyperspec-root' to point to that location."
   (interactive (list (common-lisp-hyperspec-read-symbol-name)))
-  (let ((name (common-lisp-hyperspec--strip-cl-package
-	       (downcase symbol-name)))
-        (hyperspec-root-path (let ((root (file-name-as-directory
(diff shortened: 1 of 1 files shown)


dotemacs:
> It doesn't help if the file doesn't actually exist.

Are you referring to the failing tests and their cause?

Or are you talking about the change in this PR?

It's not clear. Thanks

stassats:
Ignore the CI.

dotemacs:
Thanks for the feedback.

See the latest commit.

I tweaked the function so that it checks if the `common-lisp-hyperspec-root` is a directory and if it exists, before applying the formatting of `file://` prefix and `/` suffix, if they aren't specified.

dotemacs:
> I don't actually use the customize thingy, just a (setq common-lisp-hyperspec-root) in my .emacs. But browse-url doesn't seem to provide any way to produce an error if something doesn't exist.

It's not clear from your feedback if you'd like me to revise the latest change?

Let me know, thanks

stassats:
I don't see a way to solve this with what emacs provides.

stassats:
I would say approach 3: do nothing. How often do you specify bad urls?

stassats:
I can only accept a change that adds a missing "/" right now. The rest is a question for browse-url.

dotemacs:
> I can only accept a change that adds a missing "/" right now. The rest is a question for browse-url.

Cool, I'll add that update then. Thanks for the feedback.

stassats:
> Using if/else as it was seen as more straightforward by the
maintainer

It's not that it's not straightforward, I looked at the code of file-name-as-directory, and it does a lot of stuff. And it's not really a filename but a url.

dotemacs:
We got feedback on emacs-devel:

https://lists.gnu.org/archive/html/emacs-devel/2025-09/msg00406.html

Do you want me to address these issues?

stassats:
> It was new in Emacs-28
That's too new. 

And again, it's not a file name

PR #892, opened 2025-10-20
Restore source-path to position for SBCL's implementation of comma

csrhodes:
The source path behaves as if the comma were implemented as a two-element list (COMMA <expr>), so install a read macro to make it so.

(depends for full workingness on change 35b52dd65045b856e7e406ab79bfb4b10050a7d9, which I have just pushed to SBCL)

### DIFF
--- a/slime-tests.el
+++ b/slime-tests.el
@@ -717,10 +717,8 @@ Confirm that SUBFORM is correctly located."
        (/ 1 0)))
   (slime-test--compile-defun program subform))
 
-;; SBCL used to pass this one but since they changed the
-;; backquote/unquote reader it fails.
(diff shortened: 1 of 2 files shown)


csrhodes:
Thanks, @luismbo .  @stassats , it's probable that I've failed to account for something.

csrhodes:
I've reworked and pushed the change back into sbcl.  Assuming that it passes muster this time in sbcl itself (which I hope it will, it has a bunch of self-consistency checks, and fixes things in sb-cover, also now with tests), can we work to get slime updated in parallel?  I don't have a good story for new sbcl+old slime, unfortunately.

PR #902, opened 2026-02-12
SBCL: in definition-source-file-location, character-offset is first

digikar99:
This is incredibly silly, but it works for what I want to do - 

I'm trying to add `slime-edit-definition` support to [a non-lispy layer that transpiles to Common Lisp](https://moonli-lang.github.io/docs/intro-lisp/). The current transpilation process creates a .lisp file from a .moonli file and then the .lisp file is processed as usual. Unfortunately, this means, `slime-edit-definition` will go to the .lisp file. However, I want to send it to the appropriate place in .moonli file. 

I can obtain character-offsets in my transpiler. However, currently SLIME and SBCL insist on navigating using sexpressions, even if character-offsets are recorded. This PR first makes `swank/sbcl::definition-source-file-location` first check the character-offset instead of computing it from the form-path and sexpressions.. I'm not sure if there's a better way to achieve what I want to do (use SLIME seamlessly between lisp and moonli files).

`make check` does not break any tests. But let me know if this might cause some other issues!

### DIFF
--- a/swank/sbcl.lisp
+++ b/swank/sbcl.lisp
@@ -869,7 +869,8 @@ QUALITIES is an alist with (quality . value)"
   (with-definition-source (pathname form-path character-offset plist
                            file-write-date) definition-source
     (let* ((namestring (namestring (translate-logical-pathname pathname)))
-           (pos (or (and form-path
+           (pos (or character-offset
+                    (and form-path
                          (or
                           #+#.(swank/backend:with-symbol 'definition-source-form-number 'sb-introspect)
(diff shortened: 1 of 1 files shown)


stassats:
That's probably not good, as things move in files and using form paths is more reliable.

digikar99:
In principle, my goals can be achieved by customizing the value of `slime-find-definitions-function`. I can also rely on [definitions](https://codeberg.org/shinmera/definitions). In practice, this is going to be brittle and experimental for a while; therefore, not something that should go into slime in the near future.

PR #906, opened 2026-02-22
Fix indenting first form in the buffer

phoe:
Turns out #905 introduced a new bug, indenting the form at the beginning of the file doesn't work anymore.


phoe:
Do you know how to fix it properly? Elisp is a completely foreign land for me and I'm worried about breaking it further.

PR #907, opened 2026-02-24
SBCL: use weak hashtables for thread accounting

attila-lendvai:
i've looked into reviving some of my bitrotten commits from ages ago that were rejected during Helmut's mantainership, when he revoked my commti bit.

i'll progress in the least controversial order, at least as much as the dependencies allow reordering the commits.

this one also serves as testing the waters: any maintainer with around? is this welcome this time around?

### DIFF
--- a/swank/sbcl.lisp
+++ b/swank/sbcl.lisp
@@ -1610,42 +1610,28 @@ stack."
       (sb-thread:with-mutex (*thread-id-counter-lock*)
         (incf *thread-id-counter*)))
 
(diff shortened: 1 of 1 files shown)


attila-lendvai:
oh well... i have submitted this at least twice already, both PR's still open:

https://github.com/slime/slime/pull/237
https://github.com/slime/slime/pull/293

i should have searched before i opened this one... :)

i guess i'll wait for some feedback before i proceed much further.

PR #921, opened 2026-03-29
Fixed one compilation error & some warnings

c-kloukinas:
slime.el line 1176, slime-recompile-bytecode was calling byte-compile-file with two arguments when it only takes one.
  Its doc string states that it reloads slime but it doesn't (only byte-compiles it, which doesn't do any loading as far as its documentation goes), so I've added a load there too.

contrib/slime-tramp.el was trying to call tramp-make-tramp-file-name with different number of arguments, depending on the Emacs version, which was causing compilation warnings.
To get rid of the warnings, I've moved the treatment of the arguments outside in a macro, and now call apply on tramp-make-tramp-file-name with the argument list as derived by the macro.

Test results on Emacs 30.2 below - 4 fail in other areas (as far as I can tell).
````
Selector: t
Passed:  206
Failed:  4 (4 unexpected)
Skipped: 1
Total:   211/211

Started at:   2026-03-29 16:46:15+0100
Finished.
Finished at:  2026-03-29 16:46:50+0100

..................................F....FFF.......................s.................................................................................................................................................

F display-region-12
    For input (1 (+h 1) (+h …, test ‘slime-display-region’.
    (cl-assertion-failed (= (l2p expected-point) (point)))

F display-region-6
    For input ((+h 2) (+h 7)…, test ‘slime-display-region’.
    (cl-assertion-failed (= (l2p expected-window-start) (window-start)))

F display-region-7
    For input ((+h -2) (+h (…, test ‘slime-display-region’.
    (cl-assertion-failed (= (l2p expected-window-start) (window-start)))

F display-region-8
    For input (2 (+h 1) 3 1 …, test ‘slime-display-region’.
    (cl-assertion-failed (= (l2p expected-window-start) (window-start)))
````

Emacs 29.3 (default on Ubuntu 24.04.4 LTS Noble Numbat) chokes early on during the tests, complaining that `sldb-inspect-frame-function` (slime.el line 6024) doesn't have a docstring. It doesn't have one indeed - no idea what its contents should be.

### DIFF
--- a/contrib/slime-tramp.el
+++ b/contrib/slime-tramp.el
@@ -55,34 +55,34 @@ See also `slime-create-filename-translator'."
                            slime-filename-translations)))
         (t (list #'identity #'identity))))
 
-(defmacro tramp-make-tramp-file-name-args (username remote-host lisp-filename)
+(defmacro compat-tramp-make-tramp-file-name (username remote-host lisp-filename)
   (cond
    ((>= emacs-major-version 29)
-    `(list (tramp-find-method nil ,username ,remote-host)
-           (concat ,username
-                   "@"
-                   ,remote-host
-                   ":"
-                   ,lisp-filename)))
+    `(tramp-make-tramp-file-name (tramp-find-method nil ,username ,remote-host)
+                                 (concat ,username
+                                         "@"
+                                         ,remote-host
+                                         ":"
+                                         ,lisp-filename)))
    ((and (>= emacs-major-version 26) (< emacs-major-version 29))
     ;; Emacs 26 requires the method to be provided and the signature of
     ;; `tramp-make-tramp-file-name' has changed.
-    `(list (tramp-find-method nil ,username ,remote-host)
-           ,username
-           nil
-           ,remote-host
-           nil
-           ,lisp-filename))
+    `(tramp-make-tramp-file-name (tramp-find-method nil ,username ,remote-host)
+                                 ,username
+                                 nil
+                                 ,remote-host
+                                 nil
+                                 ,lisp-filename))
    ((boundp 'tramp-multi-methods)
-    `(list nil nil
-           ,username
-           ,remote-host
-           ,lisp-filename))
+    `(tramp-make-tramp-file-name nil nil
+                                 ,username
+                                 ,remote-host
+                                 ,lisp-filename))
    (t
-    `(list nil
-           ,username
-           ,remote-host
-           ,lisp-filename))))
+    `(tramp-make-tramp-file-name nil
+                                 ,username
+                                 ,remote-host
+                                 ,lisp-filename))))
 
 (defun slime-make-tramp-file-name (username remote-host lisp-filename)
   "Tramp compatability function.
@@ -91,8 +91,7 @@ Handles the signature of `tramp-make-tramp-file-name' changing
 over time."
   ;; Use the tramp-make-tramp-file-name-args macro, to avoid
   ;; compilation warnings.
-  (apply #'tramp-make-tramp-file-name
-         (tramp-make-tramp-file-name-args username remote-host lisp-filename)))
+  (compat-tramp-make-tramp-file-name username remote-host lisp-filename))
 
 (cl-defun slime-create-filename-translator (&key machine-instance
                                                  remote-host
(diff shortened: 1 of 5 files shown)

### CODE REVIEW

**Comment by stassats** [comment]
There's a lot of whitespace changes, making it hard to read. I'm not against whitespace changes, but not everything in the same commit. 
byte-compile-file does actually accept two arguments, but emacs, in its wisdom, likes to break things, and the second argument is now deprecated.

File: slime.el [hunk 1]
```diff
@@ -1176,9 +1176,11 @@ DIRECTORY change to this directory before starting the process.
...
   "Recompile and reload slime."
   (interactive)
-  (let ((sourcefile (concat (file-name-sans-extension (locate-library "slime"))
-                            ".el")))
-    (byte-compile-file sourcefile t)))
+  (let* ((slime-lib (file-name-sans-extension (locate-library "slime")))
+         (sourcefile (concat slime-lib
+                             ".el")))
+    (byte-compile-file sourcefile)
+    (load slime-lib)))
```
stassats: Should probably be (load (byte-compile-dest-file sourcefile))

contrib/slime-tramp.el
```diff
@@ -55,31 +55,44 @@ See also `slime-create-filename-translator'."
...
+
+(defun slime-make-tramp-file-name (username remote-host lisp-filename)
+  "Tramp compatability function.
+
+Handles the signature of `tramp-make-tramp-file-name' changing
+over time."
+  ;; Use the tramp-make-tramp-file-name-args macro, to avoid
+  ;; compilation warnings.
+  (apply #'tramp-make-tramp-file-name
+         (tramp-make-tramp-file-name-args username remote-host lisp-filename)))
```
stassats: If you are calling apply does it really need a macro?

slime.el
c-kloukinas: [byte-compile-file returs t/nil](https://www.gnu.org/software/emacs/manual/html_node/elisp/Compilation-Functions.html), so don't think that it'll work to try to load its result.

Apologies for the whitespace changes!!! I've set up my Emacs to replace tabs with spaces and delete end of line spaces before saving, so as to avoid extra diffs in my code - my genius plan clearly backfired here...

stassats: I'm suggesting byte-compile-_dest_-file

contrib/slime-tramp.el
c-kloukinas: I chose a macro, so that the cond would be evaluated just once, and not each time the function is called. Happy to change it if you don't like it.

stassats: I don't think there's a bottleneck in calling that function.

c-kloukinas: Maybe the macro should just return the full `tramp-make-tramp-file-name` call and not just its arguments, to avoid the use of apply.

stassats: We should strive for simplicity here. And actually the simplest thing would be to use https://github.com/emacs-mirror/emacs/blob/8c481ac9441ec8bc84163208a2412e9b45a96afe/lisp/subr.el#L301

slime.el
c-kloukinas: My bad, missread it. Didn't know about byte-compile-dest-file, not in the [elisp manual index](https://www.gnu.org/software/emacs/manual/html_node/elisp/Index.html).
Its description (in Emacs 30.2) mentions nothing about re-loading the file.
````
byte-compile-dest-file is a byte-code-function in ‘bytecomp.el’.

(byte-compile-dest-file FILENAME)

Convert an Emacs Lisp source file name to a compiled file name.
If ‘byte-compile-dest-file-function’ is non-nil, uses that
function to do the work.  Otherwise, if FILENAME matches
‘emacs-lisp-file-regexp’ (by default, files with the extension ".el"),
replaces the matching part (and anything after it) with ".elc";
otherwise adds ".elc".

  Probably introduced at or before Emacs version 23.2.
````

stassats: That's why (**load** (byte-compile-_dest_-file sourcefile))

c-kloukinas: OK, now I got you, sorry for being a bit slow. Changed it.

contrib/slime-tramp.el
c-kloukinas: I like your changes, removed my patch there.

stassats: I suppose you need to rebase and force push without the intermediate commits.

PR #922, opened 2026-04-21
Reload slime after bytecompiling it, as function name suggests.

c-kloukinas:
Modified function `slime-recompile-bytecode` so that it also loads the bytecompiled file, as its name suggests it should be doing.

### DIFF
--- a/slime.el
+++ b/slime.el
@@ -1186,7 +1186,8 @@ DIRECTORY change to this directory before starting the process.
   (interactive)
   (let ((sourcefile (concat (file-name-sans-extension (locate-library "slime"))
                             ".el")))
-    (byte-compile-file sourcefile t)))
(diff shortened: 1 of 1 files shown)


stassats:
The change is good, but the description is not fitting. It already does load, via the second argument to byte-compile-file. It just happens to be deprecated.

c-kloukinas:
Hi, hadn't realized that the second argument enabled that behaviour.
But if the second argument is deprecated, should we be depending on it or change it to future-proof it?

stassats:
I'm agreeing with the change, not the commit message.

c-kloukinas:
Hi, I've amended the commit message to point out that the loading now happens without using the second, deprecated argument to byte-compile-file.

PR #926, opened 2026-05-11
sbcl: Remove use of named lambda.

karlosz:
In the future, sbcl may make it so that #' does not accept special syntax anymore, in favor of some other more code-walker friendly mechanism (such as declarations). In addition, strings are generally better for just debug names to maintain consistency with FDEFINITION.

But there is actually no convincing reason to rely on sbcl internals here, so just remove the use of named lambda.

### DIFF
--- a/swank/sbcl.lisp
+++ b/swank/sbcl.lisp
@@ -1088,11 +1088,10 @@ Return a list of the form (NAME LOCATION)."
 
 (defun make-invoke-debugger-hook (hook)
   (when hook
-    #'(sb-int:named-lambda swank-invoke-debugger-hook
-          (condition old-hook)
-        (if *debugger-hook*
-            nil         ; decline, *DEBUGGER-HOOK* will be tried next.
-            (funcall hook condition old-hook)))))
+    (lambda (condition old-hook)
+      (if *debugger-hook*
+          nil         ; decline, *DEBUGGER-HOOK* will be tried next.
+          (funcall hook condition old-hook)))))
 
 (defun set-break-hook (hook)
   (setq sb-ext:*invoke-debugger-hook* (make-invoke-debugger-hook hook)))


stassats:
I don't see sb-int:named-lambda being removed. And this change doesn't even it replace it with anything.

karlosz:
I have some sbcl-side changes that would make `named-lambda` just an ordinary macro that inserts a declaration instead similar to how lambda lists are configured, which is easier for portable code walkers to accept by removing the special syntax `#'(named-lambda ...)`, so that would be a strict improvement. Would you accept a change that just removes the sharp sign and changes the symbol to a string? 

I just don't see why slime needs to reference an sbcl internal function here, it doesn't seem to add anything.

stassats:
Slime is like full of internal functions.
Which code walkers have trouble with #'named-lambda, and with #'named-lambda specifically in slime?
Is it easier to change the code-walkers than scavenge all uses of #'named-lambda?

karlosz:
We have an internal issue about it as well: https://bugs.launchpad.net/sbcl/+bug/310097

It's just used once in slime, so I was hoping we could patch it up. I hope that there isn't a lot of people doing `#'(sb-int:named-lambda ...)`.

stassats:
Changing what sb-int:named-lambda does shouldn't mean that #'(named-lambda) stops working, though.

karlosz:
I think it's more correct to disallow it eventually. We usually reject code that is not portable (in this case, it affects portable code walkers).

stassats:
Changing it after 23 years is not good, especially of little consequence. Emacs already does enough annoying things like this.

karlosz:
I understand where you are coming from, and I think it's something that can be discussed on the sbcl side further, but note that it does have consequence in that it it makes it more possible to write portable code walkers or additional macro facilities that assume that the function special form cannot have other things in them, which is not an unheard thing to do in user libraries. The bug has been open for 18 years. (Note that there are also many special cases in sbcl for handling named lambda in such cases, while under a declaration approach, no such difficulty is present.)

In this case, I believe that slime has no reason to use named lambda in the first place, and certainly not with #'. It will not break anything to make this trivial change, and only makes it easier to correct this issue in the future with less friction. So, even if one disagrees with the motivation, I don't think that should block making the harmless slime side change.

PR #929, opened 2026-06-06
Make slime-search-buffer-package ignore comments and strings

melisgl:
Previously, we relied the regexp that matches only indented in-package forms, so these didn't match:

    ;; (in-package :garbage)
    "(in-package :noise)"

However, it unfortunately matched

    #|
    (in-package :garbage)
   |#

and

   "Consider this example:

       (in-package :cl-user)
       ..."

Not anymore.

Also, I don't think it is reasonable for slime-search-buffer-package to search forwards for an in-package if it cannot find one searching backwards, but there are too many call sites for me to check, so I only added a comment about it.

### DIFF
--- a/slime-tests.el
+++ b/slime-tests.el
@@ -395,6 +395,37 @@ conditions (assertions)."
                        (slime-sexp-at-point)
                        #'eq)))
 
+(def-slime-test in-package.1 (string matches)
+  "Check that slime-search-buffer-package handles funny cases."
+  '(("(in-package :correct)"
+     ((1 1 ":correct")
+      (2 21 nil)
+      (22 22 ":correct")))
+    ("(in-package \"P1\")\n(defun foo ())\n(cL:in-PACKAGE #:p2)\n(defun bar())\n"
+     ((1 1 "\"P1\"")
+      ;; This is weird but harmless. We are in the first in-package form.
+      (2 17 "#:p2")
+      (18 53 "\"P1\"")
+      (54 69 "#:p2")))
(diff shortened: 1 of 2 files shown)


stassats:
Searching forward makes perfect sense if the in-package is the first form. There are usually comments before it, -*- stuff, etc., so it's good to see the package when opening a file.

melisgl:
Thanks. Updated to reflect this.

stassats:
Is this syntax-ppss thing expensive?

melisgl:
I did a bit of reading, and it seems that syntax-ppss may reparse from the point of a change in a buffer. If that's the case, then modifying at the beginning and invoking syntax-ppss at the end should be the worst case. I opened a 6000 line lisp file and did this:

```
(defmacro measure-time (&rest body)
  "Measure the time it takes to evaluate BODY."
  `(let ((time (current-time)))
     ,@body
     (message "%.06f" (float-time (time-since time)))))

(save-excursion
  (goto-char 1)
  (insert " (xxx)")
  (goto-line 6000)
  (measure-time (syntax-ppss (point))))
```

It's consistently below 0.01s. The sanity check of removing the insert takes 0.00002s.

Going to line 3000 takes less than 0.005s. I dare say syntax-ppss is linear in the size of the file in the worst case.

But slime-search-buffer-package-package only invokes syntax-ppss when the in-package regex matches, which should be almost always near the top of the file.

stassats:
I wonder, if there's a way to cache in-package too. This 22a7e055ca577e3de7ff5b8e3e0ab7db634c728b got reverted.

melisgl:
I have not seen a case where this was too slow yet.

stassats:
It wastes CPU cycles, searching on each key press.

PR #939, opened 2026-08-28
Support XDG directories

### LINKED ISSUE #610: Follow XDG Base Directory spec
Hello, I've recently started using emacs, getting into the ecosystem, and also learning common lisp. I've been recommended this plugin, and it seems to work really well, my only problem with this is the fact that there is - seemingly - both, no way to customise where the `~/.slime` directory should be put in, preferably somewhere that complies with the XDG Base Directory spec. What I also dislike, but on a lesser level, is the fact that it doesn't do this by default.

drlkf:
closes #610

support XDG directory spec for local files

forked from @brsvh 's version, most of the credit goes to them, i just used `getenv` from `uiop` compatibility layer to avoid hand-rolled `getenv` implementation

### DIFF
--- a/swank-loader.lisp
+++ b/swank-loader.lisp
@@ -17,6 +17,8 @@
 ;;   (setq swank-loader::*fasl-directory* "/tmp/fasl/")
 ;;   (swank-loader:init)
 
+(require 'uiop)
+
 (cl:defpackage :swank-loader
   (:use :cl)
   (:export :init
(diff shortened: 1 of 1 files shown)


stassats:
Depending on uiop is a non-starter.
And it should be an optional feature, to avoid breaking existing setups.

drlkf:
> Depending on uiop is a non-starter.

where can i find the guidelines for dependencies? i didn't find anything in contributing docs or grep in the repo

stassats:
No dependencies is the guideline. You can lift uiop:getenv wholesale, for all I care.
But silently diverging configurations is a bigger problem, I don't have a reasonable suggestion there.

stassats:
Adding SLIME_USE_XDG is the only thing that comes to mind.

brsvh (commented):
@drlkf glad to see you are trying to push forward what I half-did XD

> Depending on uiop is a non-starter.

I remember that at the time I also felt that using uiop was inappropriate, but I have forgotten how I arrived at that conclusion.

I just made a casual modification and applied the patch when build slime. At the time, I felt that XDG Base Directory was optional, so I did not bother trying to create a PR.

sdilts:
This doesn't follow the spec correctly. There are fallback directories that need to be used if the relevant XDG directory doesn't exist. See https://specifications.freedesktop.org/basedir/latest/#variables.

drlkf:
removed the dependency & provided the full spec compliance, although honestly im not sure it's super useful as 99% of implementations do not support fallback directories

